PR #16, opened 2021-04-19
updating connection info secret

alavrovinfb:
connection info secret contains DSN

### DIFF
--- a/api/v1/databaseclaim_types.go
+++ b/api/v1/databaseclaim_types.go
@@ -28,12 +28,12 @@ type DatabaseClaimSpec struct {
 	// +kubebuilder:validation:MinLength=1
 
 	// Specifies an indentifier for the application using the database.
-	AppID string `json:"appId,omitempty"`
+	AppID string `json:"appId"`
 	// +kubebuilder:validation:MinLength=0
 
 	// Specifies the type of database to provision. Only postgres is supported.
 	// +optional
-	Type string `json:"type,omitempty"`
+	Type string `json:"type"`
 
 	// In most cases the AppID will match the database name. In some cases, however, we will need to provide an optional override.
 	DBNameOverride string `json:"dbNameOverride,omitempty"`
@@ -45,7 +45,7 @@ type DatabaseClaimSpec struct {
 	InstanceLabel string `json:"instanceLabel,omitempty"`
 
 	// The username that the application will use for accessing the database.
-	Username string `json:"userName,omitempty"`
+	Username string `json:"userName"`
 
 	// The optional host name where the database instance is located.
 	// If the value is omitted, then the host value from the matching InstanceLabel will be used.
@@ -61,7 +61,7 @@ type DatabaseClaimSpec struct {
 	DatabaseName string `json:"databaseName,omitempty"`
 
 	// DSN key name.
-	DSNName string `json:"dsnName,omitempty"`
+	DSNName string `json:"dsnName"`
 }
 
 // DatabaseClaimStatus defines the observed state of DatabaseClaim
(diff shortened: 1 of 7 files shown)


api/v1/databaseclaim_types.go
```diff
@@ -87,6 +90,7 @@ type DatabaseClaimConnectionInfo struct {
 	DatabaseName string `json:"databaseName,omitempty"`
 	Username     string `json:"userName,omitempty"`
 	Password     string `json:"password,omitempty"`
```
khous, line 92: Should probably remove these unused fields

pkg/dbclient/dbClient.go
```diff
@@ -56,7 +56,7 @@ func NewPostgresClient(log logr.Logger, dbType, host, port, user, password, sslm
 	}, nil
 }
 
-func ConnectionString(host, port, user, password, sslmode string) string {
+func PostgresConnectionString(host, port, user, password, sslmode string) string {
 	return fmt.Sprintf("host=%s port=%s user=%s password=%s sslmode=%s", host, port, user, password, sslmode)
```
khous, line 60: Will password generation ever create a password with an equals sign in it?

alavrovinfb: I think it possible. Do we have to avoid the equals sign in the passwords?

mphillips-infoblox: Yup, definitely possible to have an equal sign.  https://github.com/sethvargo/go-password/blob/eb87ad851969c7b3ca007141512bd52a357a16ff/password/generate.go#L43

mphillips-infoblox: @alavrovinfb when you construct the connection string, you can single quote the values for each key (e.g. key='value'), but if you do that you will need to ensure that you escape any backslash or single quotes in the values.  See 33.1.1.1 section here: https://www.postgresql.org/docs/10/libpq-connect.html

api/v1/databaseclaim_types.go
alavrovinfb: These fields are required by HLD as part of the `DatabaseClaimConnectionInfo`, which is part of the `dbClaim`, not part of the connection secret.
Although it was decided to keep only DSN in the connection info secret, I prefer to leave the `DatabaseClaimConnectionInfo` structure as it is.
Maybe, password should be removed.
Link to HLD https://docs.google.com/document/d/1hGq3svyOx2xCuTx4qDhPut6846Y0gXp0Ne0fQpQgjcw/edit#bookmark=id.kc9cclosq6n2

alavrovinfb: @mphillips-infoblox what do think, should we update `DatabaseClaimConnectionInfo` as well as `connectionInfo` secret?

pkg/dbclient/dbClient.go
alavrovinfb: It looks like Kyle thinks the equal sign in the password can cause the problem, but it can't. Although, space in password can cause an issue but password generator doesn't use a space sign during password generation. Any way I'll add single quotes to DSN to avoid similar issues.

alavrovinfb: done

khous: I didn't see any changes to indicate that we are escaping single quotes in generate passwords. Is that not an issue?

alavrovinfb: It looks like that password generator uses a single quote sign in when generate password.
Please find full list of possible symbols https://github.com/sethvargo/go-password/blob/eb87ad851969c7b3ca007141512bd52a357a16ff/password/generate.go#L43

khous: awesome

khous (dismissed):

khous (approved):

PR #20, opened 2021-04-26
adding AWS IAM based access to RDS

alavrovinfb:


pkg/config/config.go
```diff
@@ -2,12 +2,57 @@ package config
...
+	ma.RLock()
+	defer ma.RUnlock()
+
+	return ma.authToken
+}
+
+// IsExpired checks if RDS auth toke expired
+func (ma *MasterAuth) IsExpired() bool {
+	// consider that token is expired if more than 15 - 1 minutes have passed since the token update
+	return time.Since(ma.updatedAt) > tokenExpirationTime-time.Minute
```
mphillips-infoblox, line 43: Are you able to inspect the token and extract the expiration time instead of using a hardcoded value?

alavrovinfb: As per AWS docs, RDS auth token lifetime is 15 minutes.
https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.IAMDBAuth.html

mphillips-infoblox (approved):

mphillips-infoblox (approved):

PR #31, opened 2022-04-11
Updated docs with proposed design.

seizadi:
In the case of authSource for db-controller I document existing design.


docs/index.md
```diff
@@ -227,7 +318,10 @@ DatabaseClaim:
       - [Host]: The optional host name where the database instance is located.  If the value is omitted, then the host value from the matching InstanceLabel will be used.
       - [Port]: The optional port to use for connecting to the host.  If the value is omitted, then the host value from the matching InstanceLabel will be used.
       - DatabaseName: The name of the database instance.
-
+      - Shape: The optional Shape values are arbitrary and help drive instance selection
+      - CPUs: The optional CPUs value requests the database host cpu core capacity
+      - RAM: The optional RAM value requests the database host memory capacity
```
daniel-garcia, line 323: Lets not expose RAM or CPU options to end users. They pick a shape.. nothing else.

docs/index.md
```diff
@@ -227,7 +318,10 @@ DatabaseClaim:
...
+      - Storage: The optional Storage value requests the database host storage capacity
```
daniel-garcia, line 324: This should be min-storage. The controller may over provision storage and thats ok.

docs/index.md
```diff
@@ -156,6 +245,8 @@ data:
       passwordSecretRef=athena-hostapp-password
 ```
 
+* authSource: Determines how database host master password is retrieved. The possible values are "secret" and "aws". In the case of "secret" the value from the passwordSecretRef Secret is used for the password. In the case of "aws" the RDS password is retrieved using AWS APIs and db-controller should have IAM credentials to make the necessary calls.
```
daniel-garcia, line 248: this is something the controller should decide based on its configuration

daniel-garcia (changes requested):

docs/index.md
seizadi: I am just documenting current implementation here. The new design will not use this pattern, as it is part of infrastructure concern.

docs/index.md
seizadi: OK

docs/index.md
seizadi: OK

daniel-garcia (approved):

daniel-garcia (approved):

PR #32, opened 2022-04-20
port project to kubebuilder v3

daniel-garcia:


seizadi:
I noticed that you were updating the kustomization files in config/.
For my changes I am only updating the helm charts in the project, the only dependency I have on this part is the
generate step that updates the crd definitions that are shared with helm/db-controller-crds

I noted you removed some of the Makefile targets that were tied to managing the helm chart, I assume these were intentional.

It would be good if we can get these changes reviewed so that I can merge and fix my failing unit tests.

PR #42, opened 2022-06-16
Sagar/dbproxy injector webhook

cxvisa:


config/dbproxy/dbproxysidecar.json
```diff
@@ -0,0 +1,31 @@
+{
+  "containers": [
+    {
+      "image": "docker.io/infobloxcto/sase.cloud.services:dbproxy-latest",
```
cxvisa, line 4: This image location will be changed once the dbproxy code is also pulled into this repo.  Currently only the webhook code is in this repo.

config/dbproxy/dbproxysidecar.yaml
```diff
@@ -0,0 +1,17 @@
+containers:
```
cxvisa, line 1: The above json file is actually used.  The yaml has been left as a convenience for readability.  This file will be eventually deleted.

daniel-garcia (approved):
The webhook can use a unit test but that could be addressed later.

PR #43, opened 2022-06-16
DB Proxy code from the Atlas Arch POC Repo

cxvisa:


dbproxy/pgbouncer/docker/pgbouncer.patch
```diff
@@ -0,0 +1,52 @@
+diff -uNr pgbouncer/include/bouncer.h pgbouncer.cxvisa/include/bouncer.h
+--- pgbouncer/include/bouncer.h	2022-06-13 16:56:19.178005042 -0700
++++ pgbouncer.cxvisa/include/bouncer.h	2022-06-13 16:57:54.619119705 -0700
```
daniel-garcia, line 3: nit, but this patches should probably look like 
```
a/pgbouncer/include/bouncer.h
b/pgbouncer/include/bouncer.h
``` or more aptly, not include your usersname

File: dbproxy/dbproxy.go [hunk 2]
```diff
@@ -0,0 +1,127 @@
+package main
+
+import (
+	"flag"
+	"log"
+	"os"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"infoblox.com/dbproxy/pgbouncer"
```
daniel-garcia, line 10: i think the import here should be
```gitub.com/infobloxopen/dbcontroller/dbproxy/pgbouncer```
for local builds to work properly

**Review by daniel-garcia — CHANGES_REQUESTED** [review]

File: dbproxy/dbproxy.go (hunk 2 above)
Comment (cxvisa) [reply dbproxy/dbproxy.go:10]: The local build is working fine currently.  If I make this change, compilation is failing.  The compilation is relative to dbproxy sub-directory tree.

File: dbproxy/dbproxy.go [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+	if err != nil {
+		log.Println(err)
+		panic(err)
+	}
+}
+
+func startPGBouncer() {
+	ok, err := pgbouncer.Start()
+	if !ok {
+		log.Println(err)
```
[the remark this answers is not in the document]
cxvisa: I think I will actually do os.Exit.  My intention is to indicate that there is an issue and thepod should get killed.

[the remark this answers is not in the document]
cxvisa: Goog point you brought up @daniel-garcia in our discussion -about pod not getting kicked out since there are other containers in the pod.

I will create a defect to use the livenessprobes as you suggested.

dbproxy/dbproxy.go
cxvisa: The go.mod has a replaec directive that I used to make the code compile in all scenarios (local build withe standard GOROOT and also can compile it from anywhere).

daniel-garcia (approved):

PR #50, opened 2022-06-23
Sagar/dbproxy

cxvisa:


dbproxy/pgbouncer/pgbouncer.template
```diff
@@ -11,3 +11,9 @@ pidfile = pgbouncer.pid
 admin_users = {{.UserName}}
 remote_user_override = {{.UserName}}
 remote_db_override = {{.LocalDbName}}
+client_tls_sslmode = require
```
daniel-garcia, line 14: this should be `allow` see https://www.pgbouncer.org/config.html#tls-settings client cert should not be validated.

dbproxy/start-pgbouncer.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh -x
 
+openssl req -x509 -sha256 -nodes -days 365 -newkey rsa:2048 -keyout dbproxy-server.key -out dbproxy-server.crt -subj "/C=US/CN=dbproxy-server/ST=CA/L=Santa Clara/O=Infoblox/OU=Blox in a Box"
```
daniel-garcia, line 3: create a story to use cert-manager put it in backlog. we should replace these certs with dynamically created certs that are renewed automatically

daniel-garcia (approved):

PR #65, opened 2022-07-18
Use the TAG being built in db-controller chart values

cxvisa:


Makefile
```diff
@@ -272,6 +273,7 @@ build-properties-crd:
 	@sed 's/{CHART_FILE}/$(CHART_FILE_CRD)/g' build.properties.in > crd.build.properties
 
 build-chart:
+	sed -i "/  tag: \"latest\"/c\  tag: ${TAG}" ${DB_CONTROLLER_CHART_VALUES}
```
daniel-garcia, line 276: doing this inline will lead to someone checking in the actual tag then this won't work anymore. I suggest using a .in file or following appVersion see
https://stackoverflow.com/questions/58065225/image-tag-in-values-yaml-vs-appversion-in-chart-yaml

daniel-garcia (changes requested):

Makefile
cxvisa: remove the values.yaml file from git and have values.in file that generates the values.yaml file?  Did I understand it right?

cxvisa: Added values.in file and removed values.yaml file from the git repo.  Now the values.yaml file gets generated on the fly.  We now set AppVersion in the Chart during package time (it is set to the TAG computed).  The deployment image version is picked from .Values.image.tag and if it is not set default value is set to.Chart.AppVersion.

daniel-garcia (approved):
LGTM

PR #66, opened 2022-08-29
Delete secret if DBClaim is deleted

abalaven:


controllers/databaseclaim_controller.go
```diff
@@ -818,6 +823,27 @@ func (r *DatabaseClaimReconciler) createOrUpdateSecret(ctx context.Context, dbCl
...
+	gs := &corev1.Secret{}
+	secretName := dbClaim.Spec.SecretName
+
+	err := r.Client.Get(ctx, client.ObjectKey{
+		Namespace: dbClaim.Namespace,
+		Name:      secretName,
+	}, gs)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
```
abalaven, line 836: Add warning to indicate that the object wasn't found.

daniel-garcia: This should be done by setting the owner reference. There is something incorrect about how it is being done now which is why it doesnt work. Please revisit the owner referece part of the secret creation and use the util helper methods instead of the literals that were used to create it.

abalaven: Found the root cause for the issue and made the change. Please review again.

daniel-garcia (approved):

daniel-garcia (approved):

bjeevan-ib (approved):

PR #68, opened 2022-09-15
add support for specifying sources for new DBClaims (s3bucket or existing db)

pjferrell:


api/v1/databaseclaim_types.go
```diff
@@ -23,16 +23,92 @@ import (
...
+
+// Source is a union object for specifying the initial state of a DB that should be used when provisioning a DatabaseClaim
+type Source struct {
+	// Type specifies the type of source
+	// +required
+	Type string `json:"type"`
+
+	// PostgresDatabase defines the connection information to an existing postgres db
+	// +optional
+	PostgresDatabase *PostgresDatabase `json:"postgresDatabase,omitempty"`
```
drewwells, line 55: This should be a type or an interface so we can reuse it for different types. The fields of PostgresDatabase are not unique to Postgres

drewwells: The secretRef doesn't need to be nested here either, it makes sense to be hung off top level struct

```
type: Postgres
DB: 
  DSN: "mysql://"
```

api/v1/databaseclaim_types.go
```diff
@@ -23,16 +23,92 @@ import (
...
+
+type SourceType string
+
+const (
+	PostgresSource SourceType = "postgres"
+	S3BucketSource SourceType = "s3Bucket"
+)
+
+// Source is a union object for specifying the initial state of a DB that should be used when provisioning a DatabaseClaim
+type Source struct {
```
drewwells, line 48: I'm not a fan of this name... and it could be used to define target database too. How about just `type Database` or `type DatabaseRef` ?

pjferrell: Renamed:
Source => SourceDataFrom
PostgresDatabase => Database (general purpose struct where we'll derive DB type by the dsn protocol)

api/v1/databaseclaim_types.go
pjferrell: SecretRef may have different values required in the secret.  Validating within the specific source struct will make it easier to document the requirements per source type.

drewwells (approved):

bjeevan-ib (approved):

PR #74, opened 2022-09-26
tweak some bugs in the clienter interface

drewwells:
Reworked client interface to be more extensible and remove stutter in calls ie. dbclient.DBClientFactory. Renamed interfaces to use er for go standards


pjferrell (approved):
Lgtm.

PR #75, opened 2022-09-28
Aurora support added

abalaven:
These commits incorporate the use of DB Instance and DB Cluster CRs which are used to create `aurora-postgresql` instances. RDSInstance CRs do not support deploying `aurora-postgresql` instances. For backward compatibility, RDSInstance CRs will continue to be supported.


pjferrell (dismissed):
As discussed, the controller does not handle a transition from RDSInstance based dbclaims to DBInstance based.  This requires us to ensure no dbclaims are present when switching to this version (i.e. only green field / new clusters)
We'll need to look into how we can quickly transition environments away from the older deployment method with minimal interruption.  Maintaining these different code paths will be a burden on feature development in the future.

bjeevan-ib (approved):

PR #103, opened 2022-11-17
fixed port being set as zero when provisioning aurora postgres

bjeevan-ib:


controllers/databaseclaim_controller.go
```diff
@@ -169,9 +189,54 @@ func (r *DatabaseClaimReconciler) setReqInfo(dbClaim *persistancev1.DatabaseClai
...
+		hostParams.Engine = r.Config.GetString("defaultEngine")
+	}
+
+	if hostParams.MinStorageGB == 0 {
+		hostParams.MinStorageGB = r.Config.GetInt("defaultMinStorageGB")
+	}
+
+	if hostParams.Port == 0 {
+		hostParams.Port = int64(r.Config.GetInt("defaultMasterPort"))
+	}
```
drewwells, line 222: This is a anti-pattern (cringe) in Go . Either make the default values usable or initialize hostParams as the default values.

Instead of `hostParams := DynamicHostParms{}`, do `hostParams := GetDefaultHostParams()` then you don't need all these checks for 0 and empty string.

bjeevan-ib: there is three levels from which we get these values.
1. dbclaim
2. dbcontroller config
3. default values
let me know if there is a better way to handle this.

drewwells (approved):

PR #108, opened 2022-12-08
initial checkin to support rds version upgrade

bjeevan-ib:
*support for db version upgrade
* fixed password escape issue where some of the symbols were not getting encoded/decoded as expected
* removed pg_cron from extension temporarily as this was impacting upgrade


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/infobloxopen/db-controller
 
-go 1.17
+go 1.19
```
kd7lxl: Don't forget to update https://github.com/infobloxopen/db-controller/blob/b9cdbe510ed7f65146ca08eb5836d67b24ad36a1/Jenkinsfile#L8

kd7lxl: And https://github.com/infobloxopen/db-controller/blob/b9cdbe510ed7f65146ca08eb5836d67b24ad36a1/Dockerfile#L2

bjeevan-ib: Done.

controllers/databaseclaim_controller.go
```diff
@@ -584,6 +589,7 @@ loop:
 
 	//done with migration- switch active server to newDB
 	dbClaim.Status.ActiveDB = *dbClaim.Status.NewDB.DeepCopy()
+	dbClaim.Status.ActiveDB.Status = "ready"
```
kd7lxl, line 592: Seems like there should be a type and const for this. Like this: https://github.com/Infoblox-CTO/s3-controller/blob/fb9a5fa1fc690ed7e6ab029c38992cde7ef91cba/api/v1/bucketclaim_types.go#L65-L83

controllers/databaseclaim_controller.go
```diff
@@ -733,14 +740,22 @@ func (r *DatabaseClaimReconciler) generatePassword() (string, error) {
 	minPasswordLength := r.getMinPasswordLength()
 	complEnabled := r.isPasswordComplexity()
 
+	// Customize the list of symbols.
```
kd7lxl, line 743: I can see that this code customizes the list of symbols, but what would be helpful for the comment to answer is _why_ it's necessary to customize the list.

pkg/dbclient/client.go
```diff
@@ -142,26 +138,29 @@ func (pc *client) CreateDatabase(dbName string) (bool, error) {
...
-			pc.log.Info("created extension " + s)
-		}
 	}
 
 	return created, err
 }
 
+func (pc *client) CreateDefaultExtentions(dbName string) error {
+	db, err := pc.getDB(dbName)
+	pc.log.Info("connected to " + dbName)
```
kd7lxl, line 148: Doesn't really make sense to log "connected to" before checking if there was a connection error.

pkg/dbclient/client_test.go
```diff
@@ -98,8 +98,9 @@ func setupSqlDB(t *testing.T) *testDB {
 
 	// Exponential retry to connect to database while it is booting
 	if err := pool.Retry(func() error {
-		dbConnStr := fmt.Sprintf("host=%s port=%d user=%s dbname=postgres password=%s sslmode=disable",
-			host, port, user, pass)
+		//dbConnStr := fmt.Sprintf("host=%s port=%d user=%s dbname=postgres password=%s sslmode=disable",
+		//	host, port, user, pass)
```
kd7lxl, line 102: Why is this commented out? Can it be removed?

bjeevan-ib: done

controllers/databaseclaim_controller.go
```diff
@@ -733,14 +740,23 @@ func (r *DatabaseClaimReconciler) generatePassword() (string, error) {
...
+	// Removed \ from the default list as the encoding/decoding was treating it as an escape character
```
kd7lxl, line 744: Super helpful, thank you 👍

pkg/dbclient/client_test.go
kd7lxl: Are you going to clean it up?

bjeevan-ib: @kd7lxl yikes - was looking at the wrong file.(client.go instead of client_test.go). removed it now.

kd7lxl (approved):

PR #111, opened 2022-12-15
Added tags to specify backup policy

abalaven:


bjeevan-ib (changes requested):
Would it be helpful to create a func for  the backup logic and add appropriate unit tests for the same?

also were you able to validate that these tags are applied during an update scenario? when the rds is already present.

api/v1/databaseclaim_types.go
```diff
@@ -176,6 +176,11 @@ type DatabaseClaimSpec struct {
 	// +optional
 	MinStorageGB int `json:"minStorageGB"`
 
+	// BackupPolicy specifies the duration at which db backups are taken
+	// +optional
+	// +kubebuilder:validation:Enum=Bronze;Silver;Gold
```
kd7lxl, line 181: `Bronze`, `Silver, `Gold`? Copy/paste error?

bjeevan-ib (approved):

PR #133, opened 2023-03-02
avoid upgrade of database based on storage (change) for aurora-postgres

bjeevan-ib:
storage does not apply for aurora rds.
storage should be ignored in aurora context and should not trigger an upgrade.


pkg/hostparams/hostparams.go
```diff
@@ -28,7 +33,13 @@ type HostParams struct {
 }
 
 func (p *HostParams) String() string {
-	return fmt.Sprintf("%s-%s-%s-%s", p.Engine, p.Shape, p.EngineVersion, strconv.Itoa(p.MinStorageGB))
+	if p.Engine == defaultAuroraPostgresStr {
+		//storage does not apply to aurora db. just hardcoding it a random number
+		//why 20? that is the default storage and did not want to trigger an upgrade for any existing rds
+		return fmt.Sprintf("%s-%s-%s-%s", p.Engine, p.Shape, p.EngineVersion, "20")
```
kd7lxl, line 39: I don't really understand this fake data. Why can't it just be nil?

bjeevan-ib: it can be. he only side affect will be that it will trigger an rds migration in box-3, box-4, pse-dev.
actually come to think of it, it will be a good test for migration and less confusing long term.
let me go ahead and make it empty.

bjeevan-ib: done

kd7lxl (approved):

bjeevan-ib:
closing and reopening to retrigger PR workflow (work around for failed git actions due to git hub being down yesterday)

PR #136, opened 2023-03-17
Use tagged versions of dbproxy

abalaven:


nagakonduri (approved):

config/dbproxy/dbproxysidecar.json
```diff
@@ -1,7 +1,7 @@
 {
   "containers": [
     {
-      "image": "ghcr.io/infobloxopen/dbproxy:latest",
+      "image": "...",
```
kd7lxl, line 4: What happens if you just don't define `image`? I hate to put a bad value into the spec. I'd rather leave it out so you would get a validation error if you tried to use it without setting the correct image.

abalaven: It shouldn't matter, as the only check we're doing is to unmarshal the json. I'll remove it.

abalaven: Done

kd7lxl (approved):

PR #142, opened 2023-04-03
append -v1 to image name

mphillips-infoblox:
Updating image name to match `-v1` name suffix used for ci/cd pipeline.


Makefile
```diff
@@ -1,7 +1,7 @@
 # Docker hub repo
 REGISTRY ?= infoblox
 # image name
-IMAGE_NAME ?= db-controller
+IMAGE_NAME ?= db-controller-v1
```
kd7lxl, line 4: I get an error now:
```
% make build-chart
Error: stat helm/db-controller-v1: no such file or directory
```

kd7lxl (changes requested):
Can you clarify what problem this is solving? It seems really odd to me to put the version number in the name, then prompting append another version number to the end of that (e.g., `db-controller-v1-v0.0.1-32-g1f7123e.tgz`).

helm/db-controller-legacy-crds/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-name: db-controller-crds
+name: db-controller-crds-legacy
```
kd7lxl, line 2:
```suggestion
name: db-controller-legacy-crds
```

helm/db-controller-legacy/values.yaml
```diff
@@ -1,4 +1,4 @@
-# Default values for db-controller.
+# Default values for db-controller-v1.
```
kd7lxl, line 1: Don't need `-v1`

build.properties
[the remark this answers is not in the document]
mphillips-infoblox: done

helm/db-controller-legacy-crds/Chart.yaml
mphillips-infoblox: done.

helm/db-controller-legacy/values.yaml
mphillips-infoblox: done.

kd7lxl (approved):

PR #143, opened 2023-04-12
Add ignore-startup-parameters key to pgbouncer.ini file

abalaven:


dbproxy/pgbouncer/pgbouncer.template
```diff
@@ -11,6 +11,7 @@ pidfile = pgbouncer.pid
 admin_users = {{.UserName}}
 remote_user_override = {{.UserName}}
 remote_db_override = {{.LocalDbName}}
+ignore_startup_parameters = extra_float_digits
```
kd7lxl, line 14: Should this be configurable?

kd7lxl (approved):

dbproxy/pgbouncer/pgbouncer.template
abalaven: It can be configurable. Since this file is present on the dbproxy sidecar, we'll probably need another annotation in the main deployment to indicate whether to add this line or not.

Something like `infoblox.com/db-proxy-ignore-startup-parameters: extra_float_digits`?

kd7lxl: I like that idea, but still have the question: _should this be configurable_. If we don't know, let's merge this as-is and if there are any problems we can go back and add the annotation to control it.

PR #145, opened 2023-06-09
enable replication role based on claim request

bjeevan-ib:


pkg/dbclient/client.go
```diff
@@ -342,6 +342,84 @@ func (pc *client) UpdatePassword(username string, userPassword string) error {
...
 }
 
+func (pc *client) ManageReplicationRole(username string, enableReplicationRole bool) error {
+	var (
+		exists             bool
+		hasReplicationRole bool
+	)
+
+	err := pc.DB.QueryRow(fmt.Sprintf(`SELECT EXISTS(
+					select  1 from pg_roles where rolname = %s)`, pq.QuoteLiteral(username))).Scan(&exists)
```
daniel-garcia, line 352: Doesnt the driver support string interpolation? Why construct it ?

daniel-garcia (approved):
LGTM. I think the formatted sql can be replaced with parametrized queries.

PR #146, opened 2023-06-13
change raw sql stmt to a prepared sql stmt

bjeevan-ib:


daniel-garcia (dismissed):

abalaven (approved):

pkg/dbclient/client.go
```diff
@@ -347,27 +347,35 @@ func (pc *client) ManageReplicationRole(username string, enableReplicationRole b
 		exists             bool
 		hasReplicationRole bool
 	)
+	pgRoleStmt, err := pc.DB.Prepare("SELECT EXISTS(select  1 from pg_roles where rolname = $1)")
```
daniel-garcia, line 350: prepare is great. you can also use queryrow. eg

```go
 pc.DB.QueryRow("SELECT EXISTS(select  1 from pg_roles where rolname = $1)", username).Scan(&exists)
 ```

bjeevan-ib: prepare is better than directly calling the SQL with query parameter. Especially in this case, where we are calling the same SQL multiple times.

PR #153, opened 2023-06-23
crd changes to enable superuser

bjeevan-ib:


abalaven (dismissed):

helm/db-controller/values.yaml
```diff
@@ -140,3 +141,4 @@ controllerConfig:
   athena-shared:
     masterUsername: root
   storageType: gp3
+  supportSuperUserElevation: true
```
abalaven, line 144: How are you planning to use this value?

bjeevan-ib: enablesuperUser claim will only be considered if this value is set.
check the comment in the crd

daniel-garcia (approved):

abalaven (approved):

PR #158, opened 2023-07-18
grant db permission mandated in postgres15

bjeevan-ib:


pkg/dbclient/client.go
```diff
@@ -180,6 +180,7 @@ func (pc *client) CreateGroup(dbName, rolename string) (bool, error) {
 		pc.log.Info("creating a ROLE", "role", rolename)
 		grant := `
 			GRANT ALL PRIVILEGES ON DATABASE %s TO %s;
+			GRANT ALL ON SCHEMA public TO %s;
```
abalaven, line 183: Thanks Balaji, just curious if this would return an error in versions <15.

abalaven (approved):

pkg/dbclient/client.go
bjeevan-ib: no it will not.

PR #172, opened 2023-08-29
adding options to toggle perf insight and cloudwatch logs

ssuman2-infoblox:
tested on box-3

Values in `forProvider` set by db-controller config:
image

image

### DIFF
--- a/controllers/databaseclaim_controller.go
+++ b/controllers/databaseclaim_controller.go
@@ -79,16 +79,18 @@ const (
 type ModeEnum int
 
 type input struct {
-	FragmentKey           string
-	ManageCloudDB         bool
-	SharedDBHost          bool
-	MasterConnInfo        persistancev1.DatabaseClaimConnectionInfo
-	TempSecret            string
-	DbHostIdentifier      string
-	DbType                string
-	HostParams            hostparams.HostParams
-	EnableReplicationRole bool
-	EnableSuperUser       bool
+	FragmentKey                string
+	ManageCloudDB              bool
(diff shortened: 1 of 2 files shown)


controllers/databaseclaim_controller.go
```diff
@@ -1965,7 +1984,11 @@ func (r *DatabaseClaimReconciler) updateDBInstance(ctx context.Context, dbClaim
 	if dbClaim.Spec.Type == defaultPostgresStr {
 		params := &r.Input.HostParams
 		ms64 := int64(params.MinStorageGB)
+		enablePerfInsight := r.Input.EnablePerfInsight
+		enableCloudwatchLogsExport := r.Input.EnableCloudwatchLogsExport
 		dbInstance.Spec.ForProvider.AllocatedStorage = &ms64
+		dbInstance.Spec.ForProvider.EnablePerformanceInsights = &enablePerfInsight
+		dbInstance.Spec.ForProvider.EnableCloudwatchLogsExports = enableCloudwatchLogsExport
```
bjeevan-ib, line 1991: currently we support postgres and aurora-postgres
if enablePerformance and enableCloudwatch applies to both postgres and aurora - it will have to be moved outside of the if statement.

The allocatedStorage was inside an IF statement because it was only applicable to postgres (and not aurora)

ssuman2-infoblox: yes, it's applicable to both kinds of postgres. I've updated the PR. Thanks

bjeevan-ib (approved):

PR #173, opened 2023-09-07
added BackupRetentionPeriod config

ssuman2-infoblox:
V1 Story: [B-98504](https://www10.v1host.com/InfobloxNewV1/story.mvc/Summary?oidToken=Story%3A889790&RoomContext=TeamRoom%3A699936&concept=TeamRoom)

Added BackupRetentionPolicy flag to configure it for prod and non-prod envs

### DIFF
--- a/cmd/config/config.yaml
+++ b/cmd/config/config.yaml
@@ -22,6 +22,9 @@ defaultSkipFinalSnapshotBeforeDeletion: true
 defaultPubliclyAccessible: false
 defaultDeletionPolicy: Orphan
 defaultBackupPolicyValue: Bronze
+backupRetentionDays: 0
+enablePerfInsight: false
+enableCloudwatchLogsExport: "none"
 
 passwordConfig:
   passwordComplexity: enabled
--- a/controllers/databaseclaim_controller.go
+++ b/controllers/databaseclaim_controller.go
@@ -91,6 +91,7 @@ type input struct {
 	EnableSuperUser            bool
 	EnablePerfInsight          bool
 	EnableCloudwatchLogsExport []*string
+	BackupRetentionDays        int64
 }
 
 const (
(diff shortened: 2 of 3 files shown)


bjeevan-ib (changes requested):
should we account aurora workflow?

abalaven (commented):
What are we specifying in the `backupRetentionPeriod` field?

ssuman2-infoblox:
> should we account aurora workflow?

@bjeevan-ib I've added this config for DBCluster as well. Usually, backup retention period can't be set to 0 for aurora but the value can be managed with DBCluster.

>  What are we specifying in the backupRetentionPeriod field?

@abalaven this field is used to set the number of days for which automated backups are retained. More about this is here: https://doc.crds.dev/github.com/crossplane/provider-aws/rds.aws.crossplane.io/DBInstance/v1alpha1@v0.40.0#spec-forProvider-backupRetentionPeriod

abalaven (commented):
IIUC, the default value for `backupRetentionPeriod` is set to 0 in the values file, which I believe will be added to the configMap. Is there any provision to set the value independent of the CM? The CM only holds default values if I'm not wrong.

drewwells (changes requested):
Use a string and pass it through `time.ParseDuration()`. time.Duration is represented as int64 in nanoseconds, so this will be ridiculous to configure if somebody wants 365 days
`backup: 3.1536e16`

ssuman2-infoblox:
> Use a string and pass it through `time.ParseDuration()`. time.Duration is represented as int64 in nanoseconds, so this will be ridiculous to configure if somebody wants 365 days `backup: 3.1536e16`

@drewwells @pjferrell, as per docs it only accepts value from 0-35 (0 means disabled) and it denotes number of days. We can't set it beyond 35.
image

ssuman2-infoblox:
> IIUC, the default value for `backupRetentionPeriod` is set to 0 in the values file, which I believe will be added to the configMap. Is there any provision to set the value independent of the CM? The CM only holds default values if I'm not wrong.

@abalaven I initially tried using arguments to db-controller to pass such values but it required multiple function signature changes. Also we don't want it to be set at dbclaim level but db-controller so CM seemed a good option as it's already configured to take db-controller config as input.

Added something similar here: https://github.com/infobloxopen/db-controller/pull/172

drewwells (approved):

bjeevan-ib (approved):

PR #176, opened 2023-09-11
support Platinum backup enum

bjeevan-ib:


bmuthusamy (approved):
Approve

abalaven (commented):
The CRD needs to be re-generated with `Platinum` as an additional enum value.

https://github.com/infobloxopen/db-controller/blob/main/config/crd/bases/persistance.atlas.infoblox.com_databaseclaims.yaml#L61-L64

abalaven (approved):

PR #181, opened 2023-09-20
DO NOT APPROVE/MERGE - adding dsnexec as a sidecar [OLD]

abodhekar:
This is to add dsnexec as a side-car container. 
Whenever db-controller would see following two annotations, it would add the sidecar.

        1. infoblox.com/remote-db-dsn-secret: < some-secret-name >
        2. infoblox.com/dsnexec-config-secret: < some-secret-name >
       
**_remote-db-dsn-secret_** - In case of RDS, it would be the secret created by DBClaim. This would be mounted as a volume inside the sidecar. the dsn.txt file can be watched for changes (by configuring such through dsnexec config.yaml).

**_dsnexec-config-secret_** - This would be a secret which needs to be created by the apps. This should contain data as

> key: config.yaml
> value:  < config needed for dsnexec >

This would be mounted as a volume inside the sidecar. The path to config.yaml file is given to the dsnexec entrypoint.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -173,49 +173,48 @@ func main() {
 		os.Exit(1)
 	}
 
-	webHookServer := mgr.GetWebhookServer()
+	if enableDBProxyWebhook || enableDsnExecWebhook {
+		webHookServer := mgr.GetWebhookServer()
+		webHookServer.Port = 7443
+		webHookServer.CertDir = "./certs/"
+		if enableDBProxyWebhook {
 
-	webHookServer.Port = 7443
-	webHookServer.CertDir = "./certs/"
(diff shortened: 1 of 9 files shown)


cmd/manager/main.go
```diff
@@ -166,21 +173,22 @@ func main() {
 		os.Exit(1)
 	}
 
+	webHookServer := mgr.GetWebhookServer()
+
+	webHookServer.Port = 7443
+	webHookServer.CertDir = "./certs/"
```
abalaven, line 179: We are initializing the webhook only if either dbproxy or dsnexec is enabled. We don't need the webhook if neither of them are enabled.

abodhekar: @abalaven - tx, done

PR #186, opened 2023-09-25
adding dsnexec as a sidecar [new]

abodhekar:
This is to add dsnexec as a side-car container. 
Whenever db-controller would see following two annotations, it would add the sidecar.

        1. infoblox.com/remote-db-dsn-secret: < some-secret-name >
        2. infoblox.com/dsnexec-config-secret: < some-secret-name >
       
**_remote-db-dsn-secret_** - In case of RDS, it would be the secret created by DBClaim. This would be mounted as a volume inside the sidecar. the dsn.txt file can be watched for changes (by configuring such through dsnexec config.yaml).

**_dsnexec-config-secret_** - This would be a secret which needs to be created by the apps. This should contain data as

> key: config.yaml
> value:  < config needed for dsnexec >

This would be mounted as a volume inside the sidecar. The path to config.yaml file is given to the dsnexec entrypoint.

### DIFF
--- a/webhook/webhook-dsnexec.go
+++ b/webhook/webhook-dsnexec.go
@@ -26,12 +26,16 @@ var (
 	sidecarImageForDsnExec = os.Getenv("DSNEXEC_IMAGE")
 )
 
-func dsnExecSideCardInjectionRequired(pod *corev1.Pod) (bool, string, string) {
+func dsnExecSideCarInjectionRequired(pod *corev1.Pod) (bool, string, string) {
 	remoteDbSecretName, ok := pod.Annotations["infoblox.com/remote-db-dsn-secret"]
-	dsnExecConfigSecret, ok2 := pod.Annotations["infoblox.com/dsnexec-config-secret"]
+	if !ok {
+		dsnexecLog.Info("remote-db-dsn-secret can not be seen in the annotations.", pod.Name)
+		return false, "", ""
+	}
 
-	if !ok || !ok2 {
-		dsnexecLog.Info("either or both remote-db-dsn-secret, dsnexec-config-secret can not be seen in the annotations.", pod.Name)
+	dsnExecConfigSecret, ok := pod.Annotations["infoblox.com/dsnexec-config-secret"]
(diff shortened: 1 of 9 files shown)


webhook/webhook-dsnexec.go
```diff
@@ -0,0 +1,99 @@
...
+	decoder              *admission.Decoder
+	DsnExecSidecarConfig *Config
+}
+
+var (
+	dsnexecLog             = ctrl.Log.WithName("dsnexec-controller")
+	sidecarImageForDsnExec = os.Getenv("DSNEXEC_IMAGE")
+)
+
+func dsnExecSideCardInjectionRequired(pod *corev1.Pod) (bool, string, string) {
```
jdaripineni, line 29: type in function name? *SideCar*

jdaripineni (approved):
lgtm. have a minor comment about typo

webhook/webhook-dsnexec.go
```diff
@@ -0,0 +1,99 @@
...
+	remoteDbSecretName, ok := pod.Annotations["infoblox.com/remote-db-dsn-secret"]
+	dsnExecConfigSecret, ok2 := pod.Annotations["infoblox.com/dsnexec-config-secret"]
+
+	if !ok || !ok2 {
+		dsnexecLog.Info("either or both remote-db-dsn-secret, dsnexec-config-secret can not be seen in the annotations.", pod.Name)
+		return false, "", ""
+	}
```
abalaven, line 36: Keeping these separate might be easier to debug.

abalaven (dismissed):
Looks good, just one comment and please address @jdaripineni's comment as well.

webhook/webhook-dsnexec.go
abodhekar: @jdaripineni  - done

webhook/webhook-dsnexec.go
abodhekar: @abalaven - done

abalaven (approved):

PR #187, opened 2023-09-27
fixing injection issue

abodhekar:
There were problems injecting sidecar container. 

made two fixes - 

1. added necessary parameters  to logger as it does only take arguments in key-val pair. Without such, it fatals
2. changed service name of dsnexec  webhook to same as dbproxy. Because, there was no such service present  earlier

### DIFF
--- a/webhook/webhook-dsnexec.go
+++ b/webhook/webhook-dsnexec.go
@@ -29,24 +29,24 @@ var (
 func dsnExecSideCarInjectionRequired(pod *corev1.Pod) (bool, string, string) {
 	remoteDbSecretName, ok := pod.Annotations["infoblox.com/remote-db-dsn-secret"]
 	if !ok {
-		dsnexecLog.Info("remote-db-dsn-secret can not be seen in the annotations.", pod.Name)
+		dsnexecLog.Info("remote-db-dsn-secret can not be seen in the annotations.", pod.Name, pod.Annotations)
 		return false, "", ""
 	}
 
 	dsnExecConfigSecret, ok := pod.Annotations["infoblox.com/dsnexec-config-secret"]
 	if !ok {
-		dsnexecLog.Info("dsnexec-config-secret can not be seen in the annotations.", pod.Name)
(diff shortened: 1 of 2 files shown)


webhook/webhook-dsnexec.go
```diff
@@ -29,24 +29,24 @@ var (
...
-		dsnexecLog.Info("dsnexec-config-secret can not be seen in the annotations.", pod.Name)
+		dsnexecLog.Info("dsnexec-config-secret can not be seen in the annotations.", pod.Name, pod.Annotations)
 		return false, "", ""
 	}
 
 	alreadyInjected, err := strconv.ParseBool(pod.Annotations["infoblox.com/dsnexec-injected"])
 
 	if err == nil && alreadyInjected {
-		dsnexecLog.Info("DsnExec sidecar already injected: ", pod.Name, remoteDbSecretName, dsnExecConfigSecret)
+		dsnexecLog.Info("DsnExec sidecar already injected: ", pod.Name, remoteDbSecretName, pod.Name, dsnExecConfigSecret)
```
abalaven, line 45: Is this supposed to be `pod.Name` or `pod.Annotations`?

webhook/webhook-dsnexec.go
```diff
@@ -29,24 +29,24 @@ var (
...
 		return false, remoteDbSecretName, dsnExecConfigSecret
 	}
 
-	dsnexecLog.Info("DsnExec sidecar Injection required: ", pod.Name, remoteDbSecretName, dsnExecConfigSecret)
+	dsnexecLog.Info("DsnExec sidecar Injection required: ", pod.Name, remoteDbSecretName, pod.Name, dsnExecConfigSecret)
```
abalaven, line 49: Same as above.

webhook/webhook-dsnexec.go
abodhekar: @abalaven - done

webhook/webhook-dsnexec.go
abodhekar: @abalaven - done

abalaven (approved):

PR #189, opened 2023-10-01
adding v levels to debug logs

ssuman2-infoblox:
adding V level to debug logs so they only appear in debug mode.

Added as per controller-runtime's logging documentation [here](https://github.com/kubernetes-sigs/controller-runtime/blob/main/TMP-LOGGING.md)

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -49,7 +49,7 @@ import (
 
 var (
 	scheme   = runtime.NewScheme()
-	setupLog = ctrl.Log.WithName("setup")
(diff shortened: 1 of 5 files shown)


abalaven (commented):
Nice work! Maybe you can plug a flag in the helm chart to dynamically control the value based on the env.

ssuman2-infoblox:
> Nice work! Maybe you can plug a flag in the helm chart to dynamically control the value based on the env.

Thanks @abalaven , I updated the PR to do this. Please review.
For ref: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/log/zap@v0.14.1#Options

bjeevan-ib (approved):

PR #193, opened 2023-10-25
Support system functions

bjeevan-ib:


helm/db-controller/minikube.yaml
```diff
@@ -8,3 +8,4 @@ zapLogger:
   level: debug
 dbproxy:
   enabled: false
+env: box-3
```
abalaven, line 11: Why are we defaulting this to box-3 if this values file is only for minikube?

bjeevan-ib: Yes. just for testing purpose.

abalaven (approved):

PR #196, opened 2023-11-06
adding shareProcessNamesapce feature while injecting  sidecars

abodhekar:


abalaven (approved):

drewwells:
This has broken multiple things on the cluster. Can we revert or only apply it to pods using db controller?

PR #216, opened 2023-12-11
autoscale feature added

abodhekar:


controllers/databaseclaim_controller.go
```diff
@@ -328,6 +330,21 @@ func (r *DatabaseClaimReconciler) setReqInfo(dbClaim *persistancev1.DatabaseClai
 		r.Input.EnableReplicationRole = *dbClaim.Spec.EnableReplicationRole
 	}
 
+	if hostParams.Engine == defaultPostgresStr {
```
bjeevan-ib, line 333: would it make sense to handle hostparams vars in hostparams.go instead of here?

abodhekar: @bjeevan-ib - tx, done

bjeevan-ib (approved):

PR #224, opened 2024-01-04
update end to end tests. add support for deletion policy in crd

bjeevan-ib:
logic to handle deletion policy is still TBD
wanted to update crd first to enable testing

### DIFF
--- a/helm/db-controller-crds/crd/persistance.atlas.infoblox.com_databaseclaims.yaml
+++ b/helm/db-controller-crds/crd/persistance.atlas.infoblox.com_databaseclaims.yaml
@@ -78,6 +78,14 @@ spec:
               dbVersion:
                 description: The version of the database.
                 type: string
+              deletionPolicy:
+                description: Specifies the type of deletion policy to use when the
+                  resource is deleted. It makes a lot of sense to not set it for most
+                  cases - this will default it to Orphan based on defaultDeletionPolicy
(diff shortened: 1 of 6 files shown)


helm/db-controller-crds/crd/persistance.atlas.infoblox.com_databaseclaims.yaml
```diff
@@ -78,6 +78,14 @@ spec:
...
                 description: The version of the database.
                 type: string
+              deletionPolicy:
+                description: Specifies the type of deletion policy to use when the
+                  resource is deleted. It makes a lot of sense to not set it for most
+                  cases - this will default it to Orphan based on defaultDeletionPolicy
+                  in controllerConfig If you are setting it to Delete, you should
+                  be aware that the database will be deleted when the resource is
+                  deleted. Hope you know what you are doing.
+                type: string
```
abalaven, line 88: There's a copy of the CRD present here - https://github.com/infobloxopen/db-controller/blob/main/config/crd/bases/persistance.atlas.infoblox.com_databaseclaims.yaml.

Can we set up a symlink so that we don't need to update any CRD changes in 2 places?

bjeevan-ib: Sure...but don't want to hold up this PR..should be a separate PR

abalaven (approved):

PR #239, opened 2024-06-13
Small enhancements and refactorings

EvertonCalgarotto:
Added some comments and fixed a few entries from readme file.
Renamed some functions to be more clear on what's going to happen and what each function is supposed to do.


controllers/databaseclaim_controller.go
```diff
@@ -432,7 +435,8 @@ func (r *DatabaseClaimReconciler) createMetricsDeployment(ctx context.Context, d
 	return exporter.Apply(ctx, r.Client, cfg)
 }
 
-func (r *DatabaseClaimReconciler) updateStatus(ctx context.Context, dbClaim *persistancev1.DatabaseClaim) (ctrl.Result, error) {
+// Create, migrate or upgrade database
+func (r *DatabaseClaimReconciler) executeDbClaimRequest(ctx context.Context, dbClaim *persistancev1.DatabaseClaim) (ctrl.Result, error) {
```
drewwells, line 439: upstateStatus sure was doing a lot more than CR updates

pkg/dbclient/interface.go
```diff
@@ -6,8 +6,8 @@ type Client interface {
 	CreateDatabase(dbName string) (bool, error)
```
drewwells, line 6: Since you're in here, can you rename this to `type Clienter interface`. Interfaces are supposed to end with `er`

drewwells (approved):

pkg/dbclient/interface.go
EvertonCalgarotto: That's right for interfaces with a single method, isn't it? this one has a lot of them and I don't think it would mean much.
What do you say?

PR #250, opened 2024-07-05
PTEUDO1064 - Create users, schemas, roles and rotate password

EvertonCalgarotto:
Update DBRoleClaim to hold new properties regarding user, schemas and roles to create.

Write unit tests.


README.md
```diff
@@ -112,3 +112,17 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 
+
+### Run integration tests
+
+1. Integration tests run against server box-3. Before running your tests, you'll have to update the cluster you are running against with the version you have your changes. 
+Otherwise it'll run against an old version (see next step)
```
drewwells, line 119: The integration tests install into the cluster now using `make deploy`

README.md
```diff
@@ -112,3 +112,17 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+2. Update file dbc_end2end_test.go (line 56) and file '.id' in the root folder, with a class name of your own (let's say your username). That will be used to create a namespace and the CRDs under it.
```
drewwells, line 121: Integration tests use `make` so class name is discovered automatically now.

README.md
```diff
@@ -112,3 +112,17 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+Then run:
+    make build
+    make deploy
+
+3. if you want to run a single test at a time replace It() by Fit() in the test class, this is a focused test and only this test will run (output shows failure, but this is intended, not to fool you that everything ran successfully)
+
+4. in case the tests failed in the middle, sometimes the resources are not deleted from the AWS account. So you can go in there and delete the RDSs that were created. They can be searched for by using the namespace you set in step 2.
```
drewwells, line 128: We should make sure integration tests provision crossplane CRs that when deleted also delete the AWS resources. They probably preserve databases right now

config/crd/bases/persistance.atlas.infoblox.com_dbroleclaims.yaml
```diff
@@ -62,8 +66,11 @@ spec:
                 - name
                 - namespace
                 type: object
+              username:
+                type: string
```
drewwells, line 70: This should be optional. We don't want anybody picking their usernames. We can do something like
`{basename}_{dbroleclaimname}_[a|b]`

pkg/dbclient/client.go
```diff
@@ -374,48 +424,184 @@ func (pc *client) CreateGroup(dbName, rolename string) (bool, error) {
...
+
+	if created {
+		grantSchemaPrivileges := `
+			GRANT ALL ON SCHEMA %s TO %s;
+			GRANT CONNECT ON DATABASE %s TO %s;
+		`
+		db, err := pc.getDB(dbName)
+		if err != nil {
+			pc.log.Error(err, "could not connect to db", "database", dbName)
+			return created, err
```
drewwells, line 442:
```suggestion
			return false, fmt.Errorf("connect_db %s: %w", dbName, err)
```

pkg/dbclient/client.go
```diff
@@ -374,48 +424,184 @@ func (pc *client) CreateGroup(dbName, rolename string) (bool, error) {
...
+		}
+		defer db.Close()
+		//grant schema privileges
+		_, err = db.Exec(fmt.Sprintf(grantSchemaPrivileges, pq.QuoteIdentifier(schema), pq.QuoteIdentifier(rolename), pq.QuoteIdentifier(dbName), pq.QuoteIdentifier(rolename)))
+		if err != nil {
+			pc.log.Error(err, "could not set schema privileges to role "+rolename)
+			metrics.UsersCreatedErrors.WithLabelValues("grant error").Inc()
+			return created, err
```
drewwells, line 450:
```suggestion
			metrics.UsersCreatedErrors.WithLabelValues("grant_error").Inc()
			return false, fmt.Errorf("grant_schema_privileges %s: %w", dbName, err)
```

pkg/dbclient/client_test.go
```diff
@@ -33,62 +29,62 @@ func init() {
...
-	return sql.Open("postgres", dbConnStr)
-}
+// type testDB struct {
+// 	t        *testing.T
+// 	port     int
+// 	username string
+// 	password string
+// 	resource *dockertest.Resource
+// 	pool     *dockertest.Pool
+// }
```
drewwells, line 39: Can these tests be uncommented?

pkg/dbclient/interface.go
```diff
@@ -15,6 +20,14 @@ type Client interface {
...
 	ManageCreateRole(username string, enableCreateRole bool) error
 	ManageSystemFunctions(dbName string, functions map[string]string) error
+	//Checks if a schema exists in the database
+	SchemaExists(schemaName string) (bool, error)
+	//Checks if a usedr exists in the database
+	UserExists(userName string) (bool, error)
+	//Checks if a role exists in the database
+	RoleExists(roleName string) (bool, error)
+	CreateSchema(schemaName string) (bool, error)
+	AssignRoleToUser(username, rolename string) error
```
drewwells, line 30: Consider breaking out this into a separate interface. It's very hard to mock and write tests with interfaces that implement dozen+ methods. However, you can easily pass around an object that implements multiple interfaces.

Can you also rename this interface to `Clienter`

Creater // CreateDatabase, CreateUser, CreateSchema
Exister // SchemaExists, ... 

It's fine for Clienter to implement all of the individual interfaces
```
type Clienter interface {
  Creater
  Exister
```

pkg/dbclient/testdb.go
```diff
@@ -0,0 +1,140 @@
...
+		User:   url.UserPassword(t.username, t.password),
+		Path:   "/postgres",
+	}
+	q := u.Query()
+	q.Set("sslmode", "disable")
+	u.RawQuery = q.Encode()
+	return u.String()
+}
+
+func (t *testDB) Close() {
```
drewwells, line 40:
```suggestion
func (t *testDB) Close() {
  t.Helper()
```

drewwells: Add t.Helper() on all of these helper methods

pkg/dbuser/dbuser.go
```diff
@@ -49,7 +49,7 @@ func (dbu DBUser) GetUserB() string {
 // changes in current behavior of the func.
 func (dbu DBUser) NextUser(curUser string) string {
 
-	if dbu.userA == curUser {
+	if strings.HasSuffix(curUser, SuffixA) {
```
drewwells, line 52: `NextUser()` logic tends to be more sophisticated than this. What if a completely incorrect username comes in? Maybe this needs an error and some sanity checks, especially if you're going to drop the checking on db.userA entirely here.

pkg/roleclaim/roleclaim.go
```diff
@@ -0,0 +1,450 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
+	_ "github.com/lib/pq"
```
drewwells, line 11:
```suggestion
	"github.com/lib/pq"
```

pkg/roleclaim/roleclaim.go
```diff
@@ -0,0 +1,450 @@
...
+		if apierrors.IsConflict(err) {
+			err = nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *DbRoleClaimReconciler) copySourceSecret(ctx context.Context, sourceSecret *corev1.Secret, dbRoleClaim *v1.DbRoleClaim, newUser, newPassword string) error {
```
drewwells, line 400: Are you leaving the copySourceSecret logic as is, or upgrading them to use separate username/password combinations?

drewwells (changes requested):
Field Indexer is the big miss. The rest is more like nits on the code

README.md
EvertonCalgarotto: ok, change to this then:

`1. Integration tests run against server box-3. Before running your tests, you'll have to build and deploy to make sure you are running against the latest version.

2. Update file '.id' in the root folder, with a class name of your own (let's say your username). That will be used to create a namespace and the CRDs under it.
Then run:
    make docker-build
    make deploy`

README.md
EvertonCalgarotto: included this piece here, cause resources are deleted, it's just that they might take a long time.

`(actually it might takes a long time sometimes).`

config/crd/bases/persistance.atlas.infoblox.com_dbroleclaims.yaml
EvertonCalgarotto: You mean make it optional or remove it entirely?
I guess it is better to have this property instead of using a built up name like the one you are proposing, cause a "username" property makes it easier for the consumer of the DBController (the app teams) what to expect when using the connstring and also, it's easier for them to create the DBRole request. They won't have to think what the username will be.

internal/controller/dbroleclaim_controller.go
```diff
@@ -68,239 +51,60 @@ type DbRoleClaimReconciler struct {
...
-		log.Info("creating secret", "secret", secretName, "namespace", dbRoleClaim.Namespace)
-		return r.Client.Create(ctx, role_secret)
-	}
-
-	role_secret.Data = sourceSecretData
-	log.Info("updating secret", "secret", secretName, "namespace", dbRoleClaim.Namespace)
-	return r.Client.Update(ctx, role_secret)
-
-}
+// // SetupWithManager sets up the controller with the Manager.
```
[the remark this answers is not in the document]
EvertonCalgarotto: removed.

internal/controller/dbroleclaim_controller.go
```diff
@@ -68,239 +51,60 @@ type DbRoleClaimReconciler struct {
...
 }
 
 // SetupWithManager sets up the controller with the Manager.
 func (r *DbRoleClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &persistancev1.DbRoleClaim{}, dbClaimField, func(rawObj client.Object) []string {
-		dbRoleClaim := rawObj.(*persistancev1.DbRoleClaim)
-		return []string{dbRoleClaim.Spec.SourceDatabaseClaim.Name}
-	}); err != nil {
-		return err
-	}
```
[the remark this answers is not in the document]
EvertonCalgarotto: put it back

pkg/dbclient/client.go
```diff
@@ -105,8 +105,9 @@ func PostgresConnectionString(host, port, user, password, dbname, sslmode string
 }
 
 // PostgresURI returns a URI for a postgres connection
-func PostgresURI(addr, user, password, dbname, sslmode string) string {
-	return fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=%s", url.QueryEscape(user), url.QueryEscape(password), addr, url.QueryEscape(dbname), sslmode)
+func PostgresURI(host, port, user, password, dbname, sslmode string) string {
+	return fmt.Sprintf("%s://%s:%s@%s:%s/%s?sslmode=%s", "postgres",
+		url.QueryEscape(user), url.QueryEscape(password), host, port, url.QueryEscape(dbname), sslmode)
```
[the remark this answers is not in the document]
EvertonCalgarotto: Sure, no reason to have it as paremeter.

pkg/dbclient/interface.go
EvertonCalgarotto: Good point. Created interfaces 
Creater
Exister
Updater

README.md
EvertonCalgarotto: updated

pkg/roleclaim/roleclaim.go
EvertonCalgarotto: As is, cause if no username is provided, it means we want the normal behavior of the DBRoleClaim, that is just to copy the secret under a different name.

pkg/dbuser/dbuser.go
EvertonCalgarotto: I put a verification in the roleclaim controller to validate the format of the username, so we don't get weird names

config/crd/bases/persistance.atlas.infoblox.com_dbroleclaims.yaml
EvertonCalgarotto: ok, we talked about it in slack. I'll remove this property and use <dbclaimrolename>_user_[a|b]

pkg/basefunctions/basefunctions.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+var (
+	minRotationTime = 60 * time.Minute // rotation time in minutes
+	maxRotationTime = 1440 * time.Minute
+	maxWaitTime     = 10 * time.Minute
+	defaultPassLen  = 32
+	defaultNumDig   = 10
+	defaultNumSimb  = 10
+)
```
drewwells, line 21: Not a big fan of generic packages with names that don't indicate intent. https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common
A lot of this behavior is only applicable to one other package too. Probably better to keep it closer to that controller.

drewwells: Another article on the topic https://google.github.io/styleguide/go/best-practices.html#util-packages

drewwells (approved):

pkg/basefunctions/basefunctions.go
EvertonCalgarotto: I thought of this because these values are used in multiple places. I could instead have created a class under "pkg" folder but it doesn't look right to the structure.

PR #252, opened 2024-07-08
PTEUDO-1088: added new keys to the dbclaim

leandrorichardtoledo:
- Ticket: [PTEUDO-1088 - Minor revision Maintenance Supported](https://infoblox.atlassian.net/browse/PTEUDO-1088);
  - Added two new fields to the `DBClaim`:
    -  `AutoMinorVersionUpgrade`
    - `PreferredMaintenanceWindow`
  - Refactored the dump code to check if the OS is Mac and adjust the `sed` command;


EvertonCalgarotto (approved):

api/v1/databaseclaim_types.go
```diff
@@ -223,6 +223,34 @@ type DatabaseClaimSpec struct {
...
+	// Constraints:
+	//
+	//    * Must be in the format ddd:hh24:mi-ddd:hh24:mi.
+	//
+	//    * Days must be one of Mon | Tue | Wed | Thu | Fri | Sat | Sun.
+	//
+	//    * Must be in Universal Coordinated Time (UTC).
+	//
+	//    * Must be at least 30 minutes.
+	PreferredMaintenanceWindow *string `json:"preferredMaintenanceWindow,omitempty"`
```
EvertonCalgarotto, line 253: I'd just change it to a normal string instead of a pointer to it. It just makes it more difficult to assign values to it.

leandrorichardtoledo: In that way, we would need to convert it from `string to *string` to pass it to the cross-plane.

api/v1/databaseclaim_types.go
```diff
@@ -223,6 +223,34 @@ type DatabaseClaimSpec struct {
 	// +optional
 	// +nullable
 	Tags []Tag `json:"tags,omitempty"`
+
+	// Specifies whether minor engine upgrades are applied automatically to the
+	// DB cluster during the maintenance window. By default, minor engine upgrades
+	// are applied automatically.
+	//
+	// Valid for Cluster Type: Multi-AZ DB clusters only
+	AutoMinorVersionUpgrade *bool `json:"autoMinorVersionUpgrade,omitempty"`
```
drewwells, line 232: Setting this as a `*bool` means this is disabled by default. Did someone ask to disable free, zero downtime automatic minorversion upgrades? I would just remove this, until we have a reason to support it.

leandrorichardtoledo: Removed

drewwells (approved):

PR #255, opened 2024-07-09
remove usage of dockertest and its awfulness

drewwells:
- remove all sleeps that were making tests appear to work
    - simple retry for migration status to poll for completion
    - dont wait for postgres containers to terminate. SIGKILL and move on
    - check if network is already up, create it if not found

DEMO
Before https://jenkins-cd.inca.infoblox.com/job/infobloxopen/job/db-controller/job/main/285/console
```
ok  	github.com/infobloxopen/db-controller/pkg/pgctl	80.559s	coverage: 68.4% of statements
```
After
```
ok  	github.com/infobloxopen/db-controller/pkg/pgctl	9.925s	coverage: 69.5% of statements
```

### DIFF
--- a/pkg/pgctl/pgctl_test.go
+++ b/pkg/pgctl/pgctl_test.go
@@ -1,9 +1,10 @@
 package pgctl
 
 import (
+	"cmp"
 	"database/sql"
-	"flag"
 	"fmt"
+	"net/url"
 	"os"
 	"reflect"
 	"testing"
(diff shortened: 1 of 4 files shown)


pkg/pgctl/pgctl_test.go
```diff
@@ -1017,12 +818,20 @@ func test_delete_publication_state_Execute(t *testing.T) {
...
+				t.Fatalf("delete_publication_state.Execute() = %v, want %v", got.Id(), tt.want)
 			}
 		})
 	}
 }
+
+func changeUserInfo(dsn string, username string, password string) string {
+	u, err := url.Parse(dsn)
+	if err != nil {
+		panic(err)
```
leandrorichardtoledo, line 833: I think returning an error here and panicking in the calling step would be better.

drewwells: It's a test helper either are the same

leandrorichardtoledo (approved):

PR #256, opened 2024-07-10
PTEUDO-1096 fix more integration tests

drewwells:
- unit test updated to NOT use fragmentKey
    - more unit tests to cover validation errors
    - manageError: fetch fresh claim when writing out error
    - make all status fields including connection info optional
    - refresh claim method, but be careful using it
    - reconcile after deleting object to finalize the finalizer
    - update error status when manageclouddb fails
    - check for double manageError and only log, multiple status updates are happening
    - clean up the dbinstance CRs when the tests are done
    - fix nil pointer error in testdb
    - refresh claim before updating status

### DIFF
--- a/api/v1/databaseclaim_types.go
+++ b/api/v1/databaseclaim_types.go
@@ -297,7 +297,8 @@ type Status struct {
 	// The name of the label that was successfully matched against the fragment key names in the db-controller configMap
 	MatchedLabel string `json:"matchLabel,omitempty"`
 
-	ConnectionInfo *DatabaseClaimConnectionInfo `json:"connectionInfo"`
+	// The optional Connection information to the database.
+	ConnectionInfo *DatabaseClaimConnectionInfo `json:"connectionInfo,omitempty"`
 
 	// Version of the provisioned Database
 	DBVersion string `json:"dbversion,omitempty"`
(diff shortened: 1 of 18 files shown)


drewwells:
DEMO
Some of the roleclaim integration tests fail, so here's a filter for the updated ones that now work most of the time.
```
NODEPLOY= NAMESPACE=dwells go test -ginkgo.vv -ginkgo.focus 'AWS' -run E2E
...
Ran 4 of 12 Specs in 297.029 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 8 Skipped
PASS
ok      github.com/infobloxopen/db-controller/test/e2e  297.048s
```

api/v1/databaseclaim_types.go
```diff
@@ -297,7 +297,8 @@ type Status struct {
 	// The name of the label that was successfully matched against the fragment key names in the db-controller configMap
 	MatchedLabel string `json:"matchLabel,omitempty"`
 
-	ConnectionInfo *DatabaseClaimConnectionInfo `json:"connectionInfo"`
+	// The optional Connection information to the database.
+	ConnectionInfo *DatabaseClaimConnectionInfo `json:"connectionInfo,omitempty"`
```
EvertonCalgarotto, line 301: this one is required to have DBRoleClaims published

EvertonCalgarotto: we use this, with the pass from secret, to access the DB and create roles, schemas, etc...

EvertonCalgarotto (approved):

PR #259, opened 2024-07-15
PTEUDO-1096 support crossplane CR tests in controller

drewwells:
installs the crossplane CRs db-controller depends on into envtest
    and registers the schema. manage tagging test added to exercise
    the funcionality.


pkg/databaseclaim/databaseclaim.go
```diff
@@ -1279,25 +1279,17 @@ func (r *DatabaseClaimReconciler) getReclaimPolicy(fragmentKey string) string {
...
 
-	if dbClaim.Spec.InstanceLabel == "" {
-		return true, nil
+	if dbClaim == nil {
+		return false, fmt.Errorf("nil dbclaim")
 	}
+
 	var dbClaimList v1.DatabaseClaimList
-	if err := cli.List(ctx, &dbClaimList, client.MatchingFields{instanceLableKey: dbClaim.Spec.InstanceLabel}); err != nil {
+	if err := r.Client.List(ctx, &dbClaimList, client.MatchingFields{instanceLableKey: dbClaim.Spec.InstanceLabel}); err != nil {
```
drewwells, line 1288: Technically, this fieldkey index is no longer being tested. This was a really roundabout way to verify it though. If we think it deserves testing, a more clear test is suited here.

kd7lxl (approved):

pkg/databaseclaim/tagging.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"context"
+
+	v1 "github.com/infobloxopen/db-controller/api/v1"
+)
+
+// CanTagResources checks if there's claims matching the instance
+// label of the dbClaim. If there's only one claim, it can tag
+func CanTagResources(ctx context.Context, dbClaimList v1.DatabaseClaimList, dbClaim v1.DatabaseClaim) (bool, error) {
```
leandrorichardtoledo, line 11: We do not use the `ctx`, do we need to keep it?

PR #263, opened 2024-07-18
PTEUDO-1088: changed the hash function, refactoring, working in progress

leandrorichardtoledo:
- This is a working in progress of the ticket: [PTEUDO-1165](https://infoblox.atlassian.net/browse/PTEUDO-1165);

### DIFF
--- a/pkg/databaseclaim/databaseclaim.go
+++ b/pkg/databaseclaim/databaseclaim.go
@@ -77,6 +77,7 @@ type input struct {
 	MasterConnInfo             v1.DatabaseClaimConnectionInfo
 	TempSecret                 string
 	DbHostIdentifier           string
+	OldDbHostIdentifier        string
 	HostParams                 hostparams.HostParams
 	EnableReplicationRole      bool
 	EnableSuperUser            bool
@@ -323,6 +324,7 @@ func (r *DatabaseClaimReconciler) setReqInfo(ctx context.Context, dbClaim *v1.Da
 		}
 
 		r.Input.DbHostIdentifier = r.getDynamicHostName(dbClaim)
+		r.Input.OldDbHostIdentifier = r.getOldDynamicHostName(dbClaim)
 	}
 	if r.Config.Viper.GetBool("supportSuperUserElevation") {
 		r.Input.EnableSuperUser = *dbClaim.Spec.EnableSuperUser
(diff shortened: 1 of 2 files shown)


pkg/databaseclaim/databaseclaim.go
```diff
@@ -531,99 +533,121 @@ func (r *DatabaseClaimReconciler) postMigrationInProgress(ctx context.Context, d
...
+	case M_UseExistingDB:
+		return r.handleUseExistingDB(ctx, dbClaim, logr)
+	case M_MigrateExistingToNewDB:
+		return r.handleMigrateExistingToNewDB(ctx, dbClaim, logr)
+	case M_InitiateDBUpgrade:
+		return r.reconcileMigrateToNewDB(ctx, dbClaim)
+	case M_MigrationInProgress, M_UpgradeDBInProgress:
+		return r.reconcileMigrationInProgress(ctx, dbClaim)
+	case M_UseNewDB:
+		return r.handleUseNewDB(ctx, dbClaim, logr)
```
drewwells, line 560: I like a good refactor, but did you plan on adding tests to make sure you didn't break any of this code? The only migration test we have doesn't use the db-controller at all and just calls code inside `pkg/pgctl`

pkg/databaseclaim/databaseclaim.go
```diff
@@ -531,99 +533,121 @@ func (r *DatabaseClaimReconciler) postMigrationInProgress(ctx context.Context, d
...
-			return r.manageError(ctx, dbClaim, err)
-		}
-		dbClaim.Status.ActiveDB = *dbClaim.Status.NewDB.DeepCopy()
-		dbClaim.Status.NewDB = v1.Status{ConnectionInfo: &v1.DatabaseClaimConnectionInfo{}}
+}
 
-		logr.Info("existing db reconcile complete")
-		return r.manageSuccess(ctx, dbClaim)
+// ensureConnectionInfoInitialized ensures that the ConnectionInfo is initialized.
+func ensureConnectionInfoInitialized(status *v1.Status) {
```
drewwells, line 568: this isn't a typical naming convention for initializing a pointer. `new(type)` is the way to do this or a func named `NewType` ie. `v1.NewStatus`. 

The go convention is to not need any initialization. For example,`sync.RWMutex{}` just works as is and will even throw `go vet` errors if you pass it around incorrectly. We can do the same here by removing the unnecessary pointers requiring all this initialization code.

pkg/databaseclaim/databaseclaim.go
```diff
@@ -77,6 +77,7 @@ type input struct {
 	MasterConnInfo             v1.DatabaseClaimConnectionInfo
 	TempSecret                 string
 	DbHostIdentifier           string
+	OldDbHostIdentifier        string
```
drewwells, line 80: `r.Input` is a bad idea. Reconcile is a global struct and code is attempting to store per request information in it. What happens if requestA puts a value in MasterConnInfo, then requestB doesn't setup MasterConnInfo and re-uses the one from requestA? Now, we have bugs that are based on order of requests coming in. 

I suggest any per-request parameters are stored in the context or simply derived from the `ctrl.Request` coming in. See this article on how to store things in contexts https://medium.com/@matryer/context-keys-in-go-5312346a868d

drewwells (changes requested):
I'd back out of databaseclaim changes here and focus on the new naming convention with robust testing around it. You have setup one index `0`, but we need support for more, 5 indexes. 
Name is a `[]string` with several possibilities. We may only ever use the first index `-0` but the code should be robust to be returning. The logic to find the current database needs to also consider the old static engine version, then all of the possible new names.

As for the old name, you use these variables to define it, nothing else. They are immutable now until we move everybody to new naming scheme. https://github.com/infobloxopen/db-controller/blob/main/pkg/hostparams/hostparams.go#L25-L30

PR #267, opened 2024-07-31
PTEUDO1071 - migrate new schemas users when version shape or type changes, delete CR

EvertonCalgarotto:
Migrate schemas and users when version, shape or type changes
Delete users when DBRoleClaim is deleted
Create new users and schemas when DBRoleClaim is created.
Ensure passwords are rotated.


Makefile
```diff
@@ -72,7 +72,7 @@ test-e2e: NS?=$(shell cat .id)
 test-e2e: ENV?=$(shell cat .env)
 test-e2e: .id .env
 test-e2e:
-	NAMESPACE=$(NS) ENV=$(ENV) go test ./test/e2e/ -v -ginkgo.v
+	NAMESPACE=$(NS) ENV=$(ENV) go test ./test/e2e/ -v -ginkgo.v --ginkgo.timeout=24h -timeout=1h
```
drewwells, line 75: yikes 24h 😓

pkg/databaseclaim/databaseclaim.go
```diff
@@ -372,7 +372,7 @@ func (r *DatabaseClaimReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	if permitted := isClassPermitted(r.Config.Class, *dbClaim.Spec.Class); !permitted {
 		logr.Info("ignoring this claim as this controller does not own this class", "spec", dbClaim.Spec, "controllerClass", r.Config.Class)
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: time.Hour * 2}, nil
```
drewwells, line 375: Why ever requeue this? It will always have the same error. The default is to never requeue, which is the ideal situation here.
https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/reconcile#Result

Makefile
EvertonCalgarotto: changed to 1 hour then

pkg/databaseclaim/databaseclaim.go
```diff
@@ -562,15 +563,16 @@ func (r *DatabaseClaimReconciler) executeDbClaimRequest(ctx context.Context, dbC
 	}
 	if r.mode == M_MigrateExistingToNewDB {
 		logr.Info("migrate to new  db reconcile started")
-		//check if existingDB has been already reconciled, else reconcileUseExisitngDB
+		//check if existingDB has been already reconciled, else reconcileUseExistingDB
 		existing_db_conn, err := v1.ParseUri(dbClaim.Spec.SourceDataFrom.Database.DSN)
+		logr.V(DebugLevel).Info("DSN", "M_MigrateExistingToNewDB", dbClaim.Spec.SourceDataFrom.Database.DSN)
```
drewwells, line 568: This is going to log passwords. I'd sanitize the password out first.
If there's not a helper already, something like this
```
func sanitize(dsn string) (string, error) {
  u, err := url.Parse(dbClaim.Spec.SourceDataFrom.Database.DSN)
  if err != nil {
    return "", err
  }
  info := u.UserInfo
  u.UserInfo = url.UserPassword(info.User(), "redacted")
  return u.String(), nil
}
```

pkg/databaseclaim/databaseclaim.go
```diff
@@ -803,13 +805,13 @@ func (r *DatabaseClaimReconciler) reconcileMigrateToNewDB(ctx context.Context,
 	return r.reconcileMigrationInProgress(ctx, dbClaim)
 }
 
-func (r *DatabaseClaimReconciler) reconcileMigrationInProgress(ctx context.Context,
-	dbClaim *v1.DatabaseClaim) (ctrl.Result, error) {
+func (r *DatabaseClaimReconciler) reconcileMigrationInProgress(ctx context.Context, dbClaim *v1.DatabaseClaim) (ctrl.Result, error) {
```
drewwells, line 808: Thank you, the force line wraps in method arguments really bugs me

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1250,14 +1290,14 @@ func (r *DatabaseClaimReconciler) getClientConn(dbClaim *v1.DatabaseClaim) v1.Da
...
-	logr.V(DebugLevel).Info("getting dbclient", "dsn", r.getMasterDefaultDsn())
+	logr.V(DebugLevel).Info("GET DBCLIENT", "DSN", r.getMasterDefaultDsn())
 	updateHostPortStatus(&dbClaim.Status.NewDB, r.Input.MasterConnInfo.Host, r.Input.MasterConnInfo.Port, r.Input.MasterConnInfo.SSLMode)
 	return dbclient.New(dbclient.Config{Log: log.FromContext(ctx), DBType: "postgres", DSN: r.getMasterDefaultDsn()})
 }
 
 func (r *DatabaseClaimReconciler) getMasterDefaultDsn() string {
 
-	return fmt.Sprintf("%s://%s:%s@%s:%s/%s?sslmode=%s", "postgres", r.Input.MasterConnInfo.Username, r.Input.MasterConnInfo.Password, r.Input.MasterConnInfo.Host, r.Input.MasterConnInfo.Port, "postgres", r.Input.MasterConnInfo.SSLMode)
+	return fmt.Sprintf("postgres://%s:%s@%s:%s/%s?sslmode=%s", r.Input.MasterConnInfo.Username, r.Input.MasterConnInfo.Password, r.Input.MasterConnInfo.Host, r.Input.MasterConnInfo.Port, "postgres", r.Input.MasterConnInfo.SSLMode)
```
drewwells, line 1300: Are you sure these values are properly url escaped? Passwords can have characters in them that are incompatible with URL/DSN.
 The problem with this `r.Input` is there's a big separation in the code from creating the variables and reading them raw like this. Perhaps, use a helper to read credentials from `r.Input` or better from `HostParams` ie. `r.Input.MasterDSN()`

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1285,7 +1325,7 @@ func (r *DatabaseClaimReconciler) canTagResources(ctx context.Context, dbClaim *
 	}
 
 	var dbClaimList v1.DatabaseClaimList
-	if err := r.Client.List(ctx, &dbClaimList, client.MatchingFields{instanceLableKey: dbClaim.Spec.InstanceLabel}); err != nil {
+	if err := r.Client.List(ctx, &dbClaimList, client.MatchingFields{instanceLabelKey: dbClaim.Spec.InstanceLabel}); err != nil {
```
drewwells, line 1328: This instanceLabelKey is for db-controller-legacy. `instanceLabelKey` reads properties from the config.yaml and I think ignores all fields in the CR spec itself. If you don't get any test failures, I'd just remove this indexed query on indexLabelKey.

pkg/databaseclaim/databaseclaim.go
```diff
@@ -2722,7 +2763,7 @@ func (r *DatabaseClaimReconciler) createSecret(ctx context.Context, dbClaim *v1.
 			"sslmode":        []byte(connInfo.SSLMode),
 		},
 	}
-	logr.Info("creating connection info secret", "secret", secret.Name, "namespace", secret.Namespace)
+	logr.Info("creating connection info SECRET: "+secret.Name, "secret", secret.Name, "namespace", secret.Namespace)
```
drewwells, line 2766: perhaps this, the name is duplicated in this change. Also it's easier to search for single words than when separated by spaces
```suggestion
	logr.Info("creating_connection_info", "secret", fmt.Sprintf("%s/%s", secret.Namespace, secret.Name))
```

pkg/databaseclaim/databaseclaim.go
```diff
@@ -2744,7 +2784,7 @@ func (r *DatabaseClaimReconciler) updateSecret(ctx context.Context, dsnName, dsn
 	exSecret.Data["username"] = []byte(connInfo.Username)
 	exSecret.Data["password"] = []byte(connInfo.Password)
 	exSecret.Data["sslmode"] = []byte(connInfo.SSLMode)
-	logr.Info("updating connection info secret", "secret", exSecret.Name, "namespace", exSecret.Namespace)
+	logr.Info("updating connection info SECRET: "+exSecret.Name, "secret", exSecret.Name, "namespace", exSecret.Namespace)
```
drewwells, line 2787:
```suggestion
	logr.Info("updating_connection_info", "secret", fmt.Sprintf("%s/%s", exSecret.Namespace, exSecret.Name))
```

pkg/databaseclaim/tagging_test.go
```diff
@@ -12,7 +12,7 @@ func TestCanTagResources(t *testing.T) {
 
 	claims := v1.DatabaseClaimList{
 		Items: []v1.DatabaseClaim{
-			v1.DatabaseClaim{
+			{
```
drewwells, line 15: go vet may not like this

pkg/dbclient/client.go
```diff
@@ -567,24 +567,42 @@ func (pc *client) UserExists(userName string) (bool, error) {
 	return exists, nil
 }
 
-func (pc *client) DeleteUser(username, reassignToUser string) error {
-	db := pc.DB
+func (pc *client) DeleteUser(dbName, username, reassignToUser string) error {
```
drewwells, line 570: This is a long series of async calls made into the database, can you use a context.Context and switch the calls to context versions?
```suggestion
func (pc *client) DeleteUser(ctx context.Context, dbName, username, reassignToUser string) error {
```

It makes the system work far more gracefully. We can pass a context into this client now. On shutdown, we notify the client that a shutdown is happening and it needs to stop what it is doing. Imagine if we're in the middle of a migration and we just `os.Exit(1)` on the database. The database is now in an unknown state and we may not be able to fully recovery it.

pkg/dbclient/client.go
```diff
@@ -567,24 +567,42 @@ func (pc *client) UserExists(userName string) (bool, error) {
...
 		return err
 	}
 
-	pc.log.Info("drop user", "user:", username)
-	_, err := db.Exec(fmt.Sprintf(
-		`REASSIGN OWNED BY %s TO %s;
-DROP USER %s;`, pq.QuoteIdentifier(username), pq.QuoteIdentifier(reassignToUser), pq.QuoteIdentifier(username)))
+	var exists bool
+
+	err := pc.DB.QueryRow("SELECT EXISTS(SELECT pg_user.usename FROM pg_catalog.pg_user where pg_user.usename = $1)", username).Scan(&exists)
```
drewwells, line 580: All of these calls have context variations
```suggestion
	err := pc.DB.QueryRowContext(ctx, "SELECT EXISTS(SELECT pg_user.usename FROM pg_catalog.pg_user where pg_user.usename = $1)", username).Scan(&exists)
```

pkg/dbclient/interface.go
```diff
@@ -48,7 +48,7 @@ type Exister interface {
 }
 
 type Remover interface {
-	DeleteUser(username, reassignToUser string) error
+	DeleteUser(dbName, username, reassignToUser string) error
```
drewwells, line 51:
```suggestion
	DeleteUser(ctx context.Context, dbName, username, reassignToUser string) error
```

pkg/hostparams/hostparams.go
```diff
@@ -25,7 +25,7 @@ var (
 	// DO NOT CHANGE THE DEFAULTS -
 	// They are used to determine the hash for RDS names. Any change will result in a new RDS instance being created for all applications!
 	// These values are purposely moved from the config file to the code to avoid accidental changes.
-	defaultShape         = "db.t4g.medium"
+	defaultShape         = "db.t4g.small"
```
leandrorichardtoledo, line 28: Based on the notes, it seems we should not change these values. To double-check.

pkg/roleclaim/dbroleclaim_test.go
```diff
@@ -44,6 +44,11 @@ func TestDBRoleClaimController_CreateSchemasAndRoles(t *testing.T) {
 	defer close()
 
 	viperObj.Set("passwordconfig::passwordRotationPeriod", 60)
+	viperObj.Set("defaultMasterUsername", "root")
+	viperObj.Set("defaultMasterPort", "5432")
+	viperObj.Set("defaultSslMode", "require")
+	viperObj.Set("defaultMinStorageGB", "10")
+	viperObj.Set("defaultSslMode", "disable")
```
drewwells, line 51: In general, viper has burned me when calling it inside libraries.
My suggestion in big refactors like this is to move all of the properties to a config object and read it from there. `viper` shouldn't be called outside of the main func at startup time. If we need some sort of hot reloading, that's different. We can register a method that viper can call to update properties though. It saves us from calling into viper anywhere in the package.

pkg/roleclaim/roleclaim.go
```diff
@@ -257,6 +279,182 @@ func (r *DbRoleClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	connInfo.DatabaseName = GetDBName(dbClaim)
+	return connInfo
+}
+
+func (r *DbRoleClaimReconciler) getMasterHost(dbClaim *v1.DatabaseClaim, dbcBaseConfig *dbcBaseConfig) string {
+	// If config host is overridden by db claims host
+	if dbClaim.Spec.Host != "" {
+		return dbClaim.Spec.Host
+	}
+	return r.Config.Viper.GetString(fmt.Sprintf("%s::Host", dbcBaseConfig.FragmentKey))
```
drewwells, line 331: Same comment as above, we should not rely on viper inside libraries.

pkg/roleclaim/roleclaim.go
```diff
@@ -257,6 +279,182 @@ func (r *DbRoleClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+}
+
+func (r *DbRoleClaimReconciler) getMasterUser(dbcBaseConfig *dbcBaseConfig) string {
+
+	u := r.Config.Viper.GetString(fmt.Sprintf("%s::masterUsername", dbcBaseConfig.FragmentKey))
+	if u != "" {
+		return u
+	}
+	return r.Config.Viper.GetString("defaultMasterUsername")
```
drewwells, line 340: This is what I mean about `instanceLabelKeys` aka `FragmentKey` aka total nonsense. If a fragmentkey is present, it totally ignores all of the properties of the db-controller and/or CR. If removing the fragmentkey doesn't break tests, just remove it.

pkg/roleclaim/roleclaim.go
```diff
@@ -257,6 +279,182 @@ func (r *DbRoleClaimReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+	// Find the longest prefix match
+	m, _, ok := rTree.LongestPrefix(dbClaim.Spec.InstanceLabel)
+	if !ok {
+		return "", fmt.Errorf("can't find any instance label matching fragment keys")
+	}
+
+	dbClaim.Status.ActiveDB.MatchedLabel = m
+
+	return m, nil
+}
```
drewwells, line 434: 2 birds with one stone. If you remove this method, you can also remove the `radix` package. Modern CRs doesn't rely on this `instanceLabel` work.

drewwells (commented):
Made a lot of comments. I think you can do a lot more cleanup around the instance label stuff that is dead code. There's a couple of tests that use it, but no apps on biab clusters actually use those hacks.

pkg/databaseclaim/databaseclaim.go
EvertonCalgarotto: this was in there already, but I added some url.QueryEscape in the user and pass to make it safer.

pkg/dbclient/interface.go
EvertonCalgarotto: Why is that? We don't use the context here for anything.

pkg/roleclaim/roleclaim.go
EvertonCalgarotto: AFAIK the fragmentkey is used when we want to reuse an instance and create a DB within it. If we remove this that feature won't be available anymore.

pkg/roleclaim/dbroleclaim_test.go
EvertonCalgarotto: When it's in the main code, I agree. These default values should be read when starting the app and kept somewhere in the context, to be reused when needed. But these are tests only, so I wouldn't bother too much about these in here.

pkg/roleclaim/roleclaim.go
EvertonCalgarotto: alright, I'll think of something to load these values when the app is starting and keep them in the context or somewhere more meaningful then viper itself.

pkg/hostparams/hostparams.go
EvertonCalgarotto: fixed.

pkg/databaseclaim/databaseclaim.go
EvertonCalgarotto: will remove on another task specific to this

pkg/roleclaim/roleclaim.go
EvertonCalgarotto: put all viper methods in the basefunctions class to keep them centralized and easy to find.

pkg/roleclaim/roleclaim.go
EvertonCalgarotto: will remove it on another task

drewwells (approved):

PR #273, opened 2024-08-07
Remove unused fields from claim

EvertonCalgarotto:
These fields were used in case a dbinstance was used. As this feature was removed, these fields are no longer valid.

### DIFF
--- a/pkg/databaseclaim/databaseclaim.go
+++ b/pkg/databaseclaim/databaseclaim.go
@@ -68,11 +68,8 @@ var ErrMaxNameLen = fmt.Errorf("dbclaim name is too long. max length is 44 chara
 type input struct {
 
 	// FIXME: this is type DatabaseType, not string
-	DbType string
-
-	ManageCloudDB bool
-	SharedDBHost  bool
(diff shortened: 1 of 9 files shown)


pkg/databaseclaim/databaseclaim.go
```diff
@@ -266,26 +263,6 @@ func (r *DatabaseClaimReconciler) setReqInfo(ctx context.Context, dbClaim *v1.Da
...
-	}
-	if connInfo.DatabaseName == "" {
-		return fmt.Errorf("invalid DatabaseName")
-	}
-	if strings.Contains(connInfo.DatabaseName, " ") {
-		return fmt.Errorf("invalid DatabaseName (contains space)")
-	}
-	if connInfo.Host == "" {
-		manageCloudDB = true
-	}
```
leandrorichardtoledo, line 288: How would this be set to true from now on?

EvertonCalgarotto: saw this as well running the tests. I removed that piece.

drewwells (approved):

PR #276, opened 2024-08-13
ensure only one db-controller attempts to mutate pod

drewwells:
introduce a mandatory class based label to mutate pods
    add a liveness probe to the sidecar
    get rid of these compiled min and max rotation values
    exit existing db reconciliation when nothing needs to happen
    add status predicate event filter to databaseclaim reconciler

### DIFF
--- a/internal/controller/databaseclaim_controller.go
+++ b/internal/controller/databaseclaim_controller.go
@@ -87,6 +87,7 @@ func (r *DatabaseClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	}
 
 	return ctrl.NewControllerManagedBy(mgr).
+		// Ignore status updates in objects being watched
 		WithEventFilter(predicate.GenerationChangedPredicate{}).
 		For(&persistancev1.DatabaseClaim{}).
 		Complete(r)
--- a/internal/controller/databaseclaim_controller.go
+++ b/internal/controller/databaseclaim_controller.go
@@ -18,11 +18,13 @@ package controller
 
 import (
 	"context"
+	"fmt"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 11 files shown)


internal/controller/databaseclaim_controller.go
```diff
@@ -74,7 +78,16 @@ func (r *DatabaseClaimReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if r.Config == nil {
+		return fmt.Errorf("DatabaseClaimConfig is not set")
+	}
+
+	if r.Client == nil {
+		return fmt.Errorf("client is not set")
+	}
+
 	return ctrl.NewControllerManagedBy(mgr).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
drewwells, line 90: Ignore status updates for Reconciling

pkg/databaseclaim/databaseclaim.go
```diff
@@ -571,8 +563,20 @@ func (r *DatabaseClaimReconciler) executeDbClaimRequest(ctx context.Context, dbC
...
+// bool indicates that object status should be updated
 func (r *DatabaseClaimReconciler) reconcileUseExistingDB(ctx context.Context, dbClaim *v1.DatabaseClaim) error {
-	logr := log.FromContext(ctx).WithValues("databaseclaim", dbClaim.Namespace+"/"+dbClaim.Name, "func", "reconcileUseExistingDB")
+	logr := log.FromContext(ctx).WithValues("databaseclaim", dbClaim.Namespace+"/"+dbClaim.Name)
+
+	activeDB := dbClaim.Status.ActiveDB
+	// Exit immediately if we have nothing to do
+	if activeDB.UserUpdatedAt != nil {
+		lastUpdateSince := time.Since(activeDB.UserUpdatedAt.Time)
+		if lastUpdateSince < r.getPasswordRotationTime() {
```
drewwells, line 575: Check if we need to do any user operations and exit immediately if we don't

leandrorichardtoledo (approved):

PR #286, opened 2024-08-26
PTEUDO-1247 Refactor per-request parameters to be injected into context

leandrorichardtoledo:
- It addresses the ticket: [PTEUDO-1247](https://infoblox.atlassian.net/browse/PTEUDO-1247);
- Removed input and mode global vars from the reconciler struct;
- Refactored code for passing the request info and mode when needed;


EvertonCalgarotto (approved):
Looks good!

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1355,73 +1263,70 @@ func (r *DatabaseClaimReconciler) readResourceSecret(ctx context.Context, secret
 
 // getDynamicHostName is used to name the crossplane
 // dbinstance CRs
-func (r *DatabaseClaimReconciler) getDynamicHostName(dbClaim *v1.DatabaseClaim) string {
+func (r *DatabaseClaimReconciler) getDynamicHostName(hash string, dbClaim *v1.DatabaseClaim) string {
```
drewwells, line 1266: Remove this from the reconciler object. Pass in the 'dbidentifierprefix' which is just the env name. Feel free to just rename it to envname as well.

pkg/databaseclaim/requestinfo.go
```diff
@@ -0,0 +1,95 @@
...
+	EnableReplicationRole      bool
+	EnableSuperUser            bool
+	EnablePerfInsight          bool
+	EnableCloudwatchLogsExport []*string
+	BackupRetentionDays        int64
+	CACertificateIdentifier    string
+}
+
+// setRequestInfo load base values and configs to kick off the database claim request.
+func (r *DatabaseClaimReconciler) setRequestInfo(dbClaim *v1.DatabaseClaim) (requestInfo, error) {
```
drewwells, line 27: The goal is to split all of this code up into smaller pieces. To get there, this should not hang off the reconciler struct and make a bunch of calls into the reconciler to find things. 

Change this to receive the Config and dbClaim as parameters. Made some comments above on getDynamicHostName to remove that dependence on reconciler struct too.
```suggestion
func NewRequestInfo(config *DatabaseClaimConfig, dbClaim *v1.DatabaseClaim) (requestInfo, error) {
```

drewwells (changes requested):
Some unit tests on this would be helpful to verify functionality. Made some comments to remove this from the reconciler struct so it can run independently

pkg/databaseclaim/requestinfo.go
drewwells, line 27: pass in the viper config and things to call `r.getDynamicHostName(hostParams.Hash(), dbClaim)`, so the reconciler method receiver can be removed.
`NewRequestInfo(dbClaim *v1.DatabaseClaim, ViperConfig...)`

pkg/databaseclaim/requestinfo.go
```diff
@@ -0,0 +1,95 @@
...
+		enablePerfInsight       bool
+		cloudwatchLogsExport    []*string
+		backupRetentionDays     int64
+		caCertificateIdentifier string
+	)
+
+	backupRetentionDays = basefun.GetBackupRetentionDays(r.Config.Viper)
+	caCertificateIdentifier = basefun.GetCaCertificateIdentifier(r.Config.Viper)
+	enablePerfInsight = basefun.GetEnablePerfInsight(r.Config.Viper)
+	enableCloudwatchLogsExport := basefun.GetEnableCloudwatchLogsExport(r.Config.Viper)
```
drewwells, line 39: This is a great reason to stop passing around viperconfig, it makes calling basefun methods very difficult. We should just have a config package to find these configuration parameters, then we could just do
```
backupRetentionDays = config.BackupRetentionDays
caCertificateIdentifier = config.CACertificateIdentifier
... etc
```

drewwells (approved):

PR #300, opened 2024-09-13
Remove unused vars

EvertonCalgarotto:


drewwells (approved):
My only hesitation is context should be passed to anything doing long processes or making api calls. It's used to indicate shutting down.
We close the context, wait for everybody to do their shutdowns then exit the program

PR #301, opened 2024-09-13
DEVOPS-28123 include aurora rds read replica connection information in db claim secret

EvertonCalgarotto:
Add 2 new keys into the main secret to store the replica DSN and URI_DSN

### DIFF
--- a/api/v1/databaseclaim_types.go
+++ b/api/v1/databaseclaim_types.go
@@ -28,8 +28,10 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 var (
-	DSNKey    = "dsn.txt"
-	DSNURIKey = "uri_dsn.txt"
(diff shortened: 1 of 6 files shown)


api/v1/databaseclaim_types.go
```diff
@@ -28,8 +28,10 @@ import (
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
 
 var (
-	DSNKey    = "dsn.txt"
-	DSNURIKey = "uri_dsn.txt"
+	DSNKey           = "dsn.txt"
+	DSNURIKey        = "uri_dsn.txt"
+	ReplicaDSNKey    = "ro_dsn.txt"
```
drewwells, line 33: We don't need both formats. `ro_uri_dsn.txt` is sufficient. All clients support the modern format and it's much easier to parse with `url.Parse`

drewwells (approved):

PR #302, opened 2024-09-16
DEVOPS-28230 if spec db version is empty take status db version instead if both empty throw error

EvertonCalgarotto:


pkg/databaseclaim/databaseclaim.go
```diff
@@ -1150,7 +1150,7 @@ func (r *DatabaseClaimReconciler) createDatabaseAndExtensions(ctx context.Contex
 	return nil
 }
 
-func (r *DatabaseClaimReconciler) manageUserAndExtensions(ctx context.Context, reqInfo *requestInfo, logger logr.Logger, dbClient dbclient.Clienter, status *v1.Status, dbName string, baseUsername string) error {
+func (r *DatabaseClaimReconciler) manageUserAndExtensions(reqInfo *requestInfo, logger logr.Logger, dbClient dbclient.Clienter, status *v1.Status, dbName string, baseUsername string) error {
```
drewwells, line 1153: this method is performs remote network operation(s) that should accept cancel signals via context https://go.dev/blog/context
I'd keep context on it even though the underlying clients don't currently support canceling.

Keeping that in mind, what happens if db-controller restarts while running this method? It should support context Cancel so we can cleanly shutdown this operation.

pkg/hostparams/hostparams.go
```diff
@@ -150,8 +154,13 @@ func New(config *viper.Viper, dbClaim *v1.DatabaseClaim) (*HostParams, error) {
 	}
 
 	if hostParams.EngineVersion == "" {
-		hostParams.isDefaultVersion = true
-		hostParams.EngineVersion = defaultEngineVersion
+		if dbClaim.Status.ActiveDB.DBVersion != "" {
+			log.Info("Using Status.ActiveDB.DBVersion as DefaultVersion")
```
drewwells, line 158: Not sure we want to log this, but Debug would be better. This is going to be logged a lot
```suggestion
			log.V(-1)Info("Using Status.ActiveDB.DBVersion as DefaultVersion")
```

pkg/hostparams/hostparams.go
```diff
@@ -150,8 +154,13 @@ func New(config *viper.Viper, dbClaim *v1.DatabaseClaim) (*HostParams, error) {
...
+			hostParams.EngineVersion = dbClaim.Status.ActiveDB.DBVersion
+		} else {
+			hostParams.isDefaultVersion = true
```
drewwells, line 161: Do we still need this? I also thought you were going to set this in spec.dbVersion

pkg/hostparams/hostparams.go
```diff
@@ -150,8 +154,13 @@ func New(config *viper.Viper, dbClaim *v1.DatabaseClaim) (*HostParams, error) {
 	}
 
 	if hostParams.EngineVersion == "" {
-		hostParams.isDefaultVersion = true
-		hostParams.EngineVersion = defaultEngineVersion
+		if dbClaim.Status.ActiveDB.DBVersion != "" {
+			log.Info("Using Status.ActiveDB.DBVersion as DefaultVersion")
+			hostParams.EngineVersion = dbClaim.Status.ActiveDB.DBVersion
```
drewwells, line 159: Modifies behavior of finding crossplane CR, so status is considered if it's available https://github.com/infobloxopen/db-controller/blob/625c742b7789cefb3c6e5b7934a162a9c4f13d6d/pkg/hostparams/hostparams.go#L63

pkg/hostparams/hostparams.go
EvertonCalgarotto: no need, Spec.DBVersion might cause circular updates with Flux and DBController itself....where Flux will set it to empty and dbcontroller will set it to status.dbversion

pkg/databaseclaim/databaseclaim.go
EvertonCalgarotto: I'd put it back when needed.

drewwells (approved):

PR #306, opened 2024-09-24
DEVOPS-28230 remove version validation

EvertonCalgarotto:
If NEW Claim and DBVersion="" version 15 is used instead
If EXISTING Claim and DBVersion= "" then version from Status.ActiveDB is used


internal/controller/databaseclaim_controller_test.go
```diff
@@ -131,11 +131,11 @@ var _ = Describe("DatabaseClaim Controller", func() {
 
 		})
 
-		It("Should fail to reconcile DB Claim missing dbVersion", func() {
+		It("Should succeed to reconcile DB Claim missing dbVersion", func() {
 			By("Reconciling the created resource")
 
 			_, err := controllerReconciler.Reconcile(ctx, reconcile.Request{NamespacedName: typeNamespacedName})
-			Expect(err).To(MatchError(".spec.dbVersion is a mandatory field and cannot be empty"))
+			Expect(err).NotTo(HaveOccurred())
```
drewwells, line 138: Probably better to still have some sort of error here. I believe the idea of this set of tests is to verify dbclaims update and can be fixed by changing their values. What can we do to continue to simulate that error MajorVersion `12`? Invalid database type?

pkg/databaseclaim/awsprovider.go
```diff
@@ -12,10 +12,13 @@ import (
 	_ "github.com/lib/pq"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const defaultMajorVersion string = "15"
```
drewwells, line 20: Parameterize this in the helm chart. KTLO team is going to update this version over time and they shouldn't need to recompile the software to change this version.

pkg/databaseclaim/awsprovider.go
```diff
@@ -137,9 +135,9 @@ func (r *DatabaseClaimReconciler) manageDBClusterAWS(ctx context.Context, dbHost
 						DBClusterParameterGroupNameRef: &xpv1.Reference{
 							Name: pgName,
 						},
-						EngineVersion: &params.EngineVersion,
+						EngineVersion: (map[bool]*string{true: ptr.To(defaultMajorVersion), false: &params.DBVersion})[params.IsDefaultVersion],
```
drewwells, line 138: This is clever, but confusing. Can we just validate the params.DBVersion early in the workflow? If it's empty, set params.DBVersion=defaultMajorVersion

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1116,11 +1116,11 @@ func (r *DatabaseClaimReconciler) getParameterGroupName(hostParams *hostparams.H
 
 	switch dbType {
 	case v1.Postgres:
-		return hostName + "-" + (strings.Split(hostParams.EngineVersion, "."))[0]
+		return hostName + "-" + (strings.Split(hostParams.DBVersion, "."))[0]
```
drewwells, line 1119: This is a panic waiting to happen, actually I think by default `DBVersion==15` would panic here. 

Do some sanity checks on the dbversion. Afterall, users can set this field

```
func (r *DatabaseClaimReconciler) getParameterGroupName(hostParams *hostparams.HostParams, dbClaim *v1.DatabaseClaim, dbType v1.DatabaseType) (string, error) {
    parts := strings.Split(hostParams.DBVersion, ".")
    if len(parts) != 2 {
       return "", fmt.Errorf("invalid db version found: %s", hostParams.DBVersion)
    }
   ...
```

pkg/hostparams/hostparams.go
```diff
@@ -99,27 +100,20 @@ func (p *HostParams) HasEngineChanged(activeEngine string) bool {
...
+	return activeEngine != p.Type
 }
 
 func (p *HostParams) HasVersionChanged(activeVersion string) bool {
-	if p.isDefaultVersion {
+	if p.IsDefaultVersion {
 		return false
 	}
-	return activeVersion != p.EngineVersion
+	return strings.Split(activeVersion, ".")[0] != strings.Split(p.DBVersion, ".")[0]
```
drewwells, line 110: unsafe use of indexing a slice. Maybe make a type, there's a lot of comparisons going around here

```
type Version struct {
  Major, Minor string
}
v := Version{Major:"15", "Minor: "3"}
return cmp.Equal(v.Major, "15")
```

drewwells (commented):
Some unsafe use of slices that will panic. Otherwise looks good

pkg/databaseclaim/databaseclaim.go
EvertonCalgarotto: wrote a test to show it works when it's 15 or 15.2 or 15.2.2 (invalid, but test is test)

pkg/hostparams/hostparams.go
EvertonCalgarotto: this works if the string is "", "15" or "15.2"
looks ugly, but works

drewwells (approved):

PR #324, opened 2024-10-01
PTEUDO-1562 rewrite migration logic of controller

drewwells:
The state machine was not being used effectively to go from the
    new cloud db step to the migration step. The controller now
    reconciles fully after a database is created and requeues. On
    next requeue, controller performs a migration to the new database.

    I attempted to identify issues storing, retrieiving and using db
    credentials. In all places possible, a uri dsn is now used. There
    is still significant cleanup in this regard. The usage of storing
    state in requestinfo needs to be eliminated. RequestInfo makes the
    code perform in a spaghetti fashion where you follow logic only to
    find it calling other states non-recurisvely or putting the dbc in
    a broken state when random errors occur.

    - fix grant/revoke superuser
    - support gcp alloydb roles
    - pass cloud into dbclient pkg calls
    - ignore sql revoke errors on non-cloud dbs
    - separate kubectl client for finding cloud based master credentials
    - push sql updatepassword when credentials are denied
    - always set status when existing reconcileNewDB
    - persist source user creds to secret at end of newdb for migration
    - bind psql containers to host IP so they can talk to each other
    - refactor ways credentials are divined from cluster
    - move mock sql roles into testdb runner
    - systemfunctions cleanup logic and logs
    - ignore errors during dbproxy shutdown

### DIFF
--- a/api/v1/databaseclaim_types.go
+++ b/api/v1/databaseclaim_types.go
@@ -17,6 +17,7 @@ limitations under the License.
 package v1
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 	"strings"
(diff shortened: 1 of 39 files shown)


drewwells:
Running this to see how we're doing on reliable helm tests
```
count=0; while make undeploy deploy helm-test; do ((count+=1)); done; echo "Command failed after $count successful executions"
```

PR #326, opened 2024-10-03
Changed to read db identifier prefix from system functions ib_env var

leandrorichardtoledo:


drewwells:
Can you take a look at the failed tests?

drewwells (approved):

PR #331, opened 2024-10-09
DEVOPS-28607 pass env into configmap

drewwells:


drewwells:
Verified this is working on box-3. Notice dbcluster now has an env identifier on it
```
k get dbcluster.rds box-3-example-regular-35d9a478
NAME                             READY   SYNCED   EXTERNAL-NAME                    AGE
box-3-example-regular-35d9a478   True    True     box-3-example-regular-35d9a478   81s
```

PR #336, opened 2024-10-16
PTEUDO-1616: update dsns when a new role claim is created

leandrorichardtoledo:
- https://infoblox.atlassian.net/browse/PTEUDO-1616


pkg/roleclaim/roleclaim.go
```diff
@@ -585,24 +581,53 @@ func (r *DbRoleClaimReconciler) copySourceSecret(ctx context.Context, sourceSecr
 	sourceSecretData := sourceSecret.Data
 
 	if newUser != "" {
+		oldUser := string(sourceSecretData["username"])
```
drewwells, line 584: This is a lot of duplication. How about
```
if newUser != "" || newPassword != "" {
```

pkg/roleclaim/roleclaim.go
```diff
@@ -585,24 +581,53 @@ func (r *DbRoleClaimReconciler) copySourceSecret(ctx context.Context, sourceSecr
...
+
 	if newPassword != "" {
+		oldPassword := string(sourceSecretData["password"])
+
+		dsn := string(sourceSecretData["dsn.txt"])
+		dsn = strings.Replace(dsn, "password="+oldPassword, "password="+newPassword, 1)
+		sourceSecretData["dsn.txt"] = []byte(dsn)
+
+		roDsn := string(sourceSecretData["ro_uri_dsn.txt"])
+		roDsn = strings.Replace(roDsn, oldPassword, newPassword, 1)
```
drewwells, line 609: You have to escape the passwords so they're URL safe. Please use the dbclient methods for populating DSN as they're verified to work.

https://github.com/infobloxopen/db-controller/blob/main/pkg/databaseclaim/secrets.go#L33-L34

drewwells (changes requested):
passwords need url escaping

pkg/roleclaim/roleclaim.go
leandrorichardtoledo: Refactored.

pkg/roleclaim/roleclaim.go
```diff
@@ -587,22 +583,36 @@ func (r *DbRoleClaimReconciler) copySourceSecret(ctx context.Context, sourceSecr
...
+	database := string(sourceSecretData["database"])
+	hostname := string(sourceSecretData["hostname"])
+	password := string(sourceSecretData["password"])
+	port := string(sourceSecretData["port"])
+	sslmode := string(sourceSecretData["sslmode"])
+	username := string(sourceSecretData["username"])
+
+	sourceSecretData["dsn.txt"] = []byte(dbclient.PostgresConnectionString(hostname, port, username, password, database, sslmode))
+	sourceSecretData["uri_dsn.txt"] = []byte(dbclient.PostgresURI(hostname, port, username, password, database, sslmode))
+	if sourceSecretData["ro_uri_dsn.txt"] != nil {
```
drewwells, line 600: I think default of []byte is nil, so this is always going to be nil

pkg/roleclaim/roleclaim.go
```diff
@@ -587,22 +583,36 @@ func (r *DbRoleClaimReconciler) copySourceSecret(ctx context.Context, sourceSecr
...
+		sourceSecretData["ro_uri_dsn.txt"] = []byte(dbclient.PostgresURI(hostname, port, username, password, database, sslmode))
```
drewwells, line 601: `ro_uri_dsn.txt` is different than `uri_dsn.txt` 
https://github.com/infobloxopen/db-controller/blob/main/pkg/databaseclaim/secrets.go#L37

drewwells (commented):
ro dsn needs a little work

PR #353, opened 2024-10-31
Leandro/close admin db connection

leandrorichardtoledo:


pkg/dbclient/client.go
```diff
@@ -961,11 +961,27 @@ func (pc *client) RevokeAccessToRole(username, rolename string) error {
 }
 
 func (pc *client) Close() error {
+	var err error
+
 	if pc.DB != nil {
-		return pc.DB.Close()
+		if dbErr := pc.DB.Close(); dbErr != nil {
+			pc.log.Error(dbErr, "could not close DB")
+			err = fmt.Errorf("could not close DB: %w", dbErr)
```
drewwells, line 969:
```suggestion
			errs = append(errs, fmt.Errorf("could not user client conn: %w", dbErr))
```

pkg/dbclient/client.go
```diff
@@ -961,11 +961,27 @@ func (pc *client) RevokeAccessToRole(username, rolename string) error {
...
+	}
+
+	if pc.adminDB != nil {
+		if adminErr := pc.adminDB.Close(); adminErr != nil {
+			pc.log.Error(adminErr, "could not close admin DB")
+			if err != nil {
+				err = fmt.Errorf("%v; %w", err, adminErr)
+			} else {
+				err = fmt.Errorf("could not close admin DB: %w", adminErr)
+			}
```
drewwells, line 980:
```suggestion
			errs = append(errs, adminErr)
```

pkg/dbclient/client.go
```diff
@@ -961,11 +961,27 @@ func (pc *client) RevokeAccessToRole(username, rolename string) error {
...
+		}
 	}
 
-	return fmt.Errorf("can't close nil DB")
+	return err
```
drewwells, line 984:
```suggestion
        // This will be nil if all errs are nil
	return errors.Join(errs)
```

drewwells (changes requested):
Go has some clumsy methods for list of errors

drewwells (approved):

PR #356, opened 2024-11-09
PTEUDO-1991 drop database prior to restoring it

drewwells:
When a database is partially or fully restored, we fail
    to migrate to it. To support migration to existing dbs,
    we will drop the database. This requires dropping all
    current connections to the database in the process.

    Add a migration test to a previously migrated database.


drewwells:
Dropping the database was more pain than it was worth. I think we can get by with dropping the schema BEFORE attempting a restore. The schemas will be recreated by name as the role associated with the application.

drewwells:
Kind of a lot of changes, let's do this after the 1.10.1 release

PR #362, opened 2024-11-18
PTEUDO-1658: update dashboard

leandrorichardtoledo:


drewwells (approved):

helm/db-controller/values.yaml
```diff
@@ -184,3 +184,5 @@ controllerConfig:
     ib_env: "{{ .Values.env }}"
     ib_lifecycle: "{{ .Values.lifecycle }}"
 
+dashboard:
+  enabled: true
```
drewwells, line 188:
```suggestion
  enabled: false
```
Don't want to add a dependency if it's not needed

PR #370, opened 2024-12-04
PTEUDO-2142 remove validation of DSNName

drewwells:
Previously, customizing the dsnname would throw a validation error
    and prevent any processing of the databaseclaim. This allows the
    customization but ensures the normalized fields are always
    populated.

https://infoblox.atlassian.net/browse/PTEUDO-2142

### DIFF
--- a/internal/controller/databaseclaim_controller_test.go
+++ b/internal/controller/databaseclaim_controller_test.go
@@ -33,6 +33,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	persistancev1 "github.com/infobloxopen/db-controller/api/v1"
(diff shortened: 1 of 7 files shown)


bjeevan-ib (approved):

leandrorichardtoledo (approved):

internal/controller/databaseclaim_controller_test.go
```diff
@@ -78,6 +99,8 @@ var _ = Describe("DatabaseClaim Controller", func() {
 					Namespace: "default",
 				},
 				Spec: persistancev1.DatabaseClaimSpec{
+					// TODO: remove customization of DSNName
```
daniel-garcia, line 102: We should not remove it. Just re-implement it.

PR #373, opened 2024-12-06
PTOCP-1756 check activedb.dbversion is respected

drewwells:
We store the dbversion for looking up crossplane CRs in the
    status field of the databaseclaim. This test verifies that
    the databaseclaim is able to find the version when its not
    the default one that the controller would use.

Enforces a unit test to ensure we lookup the activedb.dbversion. If this lookup is missing, 99% of our databases would fail to be resolved as seen in [PTOCP-1756](https://infoblox.atlassian.net/browse/PTOCP-1756)

ref: https://github.com/infobloxopen/db-controller/pull/369

### DIFF
--- a/pkg/hostparams/hostparams_test.go
+++ b/pkg/hostparams/hostparams_test.go
@@ -808,7 +808,7 @@ func TestNew_dbversion(t *testing.T) {
 		wantErr bool
 	}{
 		{
-			name: "ctrl_default",
+			name: "default_version",
 			claim: &persistancev1.DatabaseClaim{
 				Status: persistancev1.DatabaseClaimStatus{
 					ActiveDB: persistancev1.Status{},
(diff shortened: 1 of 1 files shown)


pkg/hostparams/hostparams_test.go
```diff
@@ -792,10 +782,68 @@ func TestDBVersions(t *testing.T) {
...
+func TestNew_dbversion(t *testing.T) {
+
+	var tests = []struct {
+		name    string
+		claim   *persistancev1.DatabaseClaim
+		want    *HostParams
+		wantErr bool
+	}{
+		{
+			name: "ctrl_default",
```
leandrorichardtoledo, line 811: both test cases have the same name

leandrorichardtoledo (approved):

PR #391, opened 2025-01-14
DEVOPS-30643: adds pgbouncer config to set the role before running SQL commands.

leandrorichardtoledo:
- This adds the [server_reset_query](https://www.pgbouncer.org/config.html#connection-sanity-checks-timeouts) configuration to pgbouncer.ini file to allow the queries to be executed with the role (spire-user) and not the user (spire-user_a);


dbproxy/pgbouncer/config.go
```diff
@@ -156,6 +96,9 @@ func parseURI(c *PGBouncerConfig, dsn string) error {
 		return fmt.Errorf("invalid_scheme: %s", u.Scheme)
 	}
 
+	// Remove the _<a|b> suffix from the username.
+	c.RoleName = strings.Split(u.User.Username(), "_")[0]
```
drewwells, line 100: Can we be more explicit here and pass in the exact role name ? 
We don't prevent developers from putting _ in their username and then this parsing would fail.

drewwells (commented):
Is there any test you can add to verify the role permission behavior?

dbproxy/pgbouncer/config.go
leandrorichardtoledo: Pushed the change.

leandrorichardtoledo:
@drewwells 
> Is there any test you can add to verify the role permission behavior?

I ran the test I described in the ticket [DEVOPS-30643](https://infoblox.atlassian.net/browse/DEVOPS-30643?focusedCommentId=4006198);
The next test I was looking for was to merge this and test on boxes 3 and 4 with the spire server.

dbproxy/pgbouncer/config.go
```diff
@@ -156,6 +96,7 @@ func parseURI(c *PGBouncerConfig, dsn string) error {
 		return fmt.Errorf("invalid_scheme: %s", u.Scheme)
 	}
 
+	c.RoleName = strings.TrimSuffix(strings.TrimSuffix(u.User.Username(), "_a"), "_b")
```
drewwells, line 99:
```suggestion
	// Ensure this maps to dbuser.SuffixA dbuser.SuffixB
	c.RoleName = strings.TrimSuffix(strings.TrimSuffix(u.User.Username(), "_a"), "_b")
```

drewwells (approved):

PR #393, opened 2025-01-15
DEVOPS-30870: fix password rotation logic

leandrorichardtoledo:
- This PR fixes the user/password rotation logic that was broken;
- It fixes it by checking if the current user is empty and activeDB status struct has a URI, if yes, this means there is an activeDB already with an active user and not an initial database creation;
- Refer to: https://infoblox.atlassian.net/browse/PTEUDO-2215?focusedCommentId=4007818

### DIFF
--- a/pkg/databaseclaim/databaseclaim.go
+++ b/pkg/databaseclaim/databaseclaim.go
@@ -1172,17 +1172,6 @@ func (r *DatabaseClaimReconciler) manageUserAndExtensions(ctx context.Context, r
 		return fmt.Errorf("newDB connection info is nil")
 	}
 
-	// The statusActiveDB is relevant only in the UseNewDB mode for now.
-	if operationalMode == M_UseNewDB {
-		if statusActiveDB == nil {
-			return fmt.Errorf("status activeDB is nil")
-		}
-
-		if statusActiveDB.ConnectionInfo == nil {
-			return fmt.Errorf("activeDB connection info is nil")
-		}
-	}
-
 	dbu := dbuser.NewDBUser(baseUsername)
 	rotationTime := r.getPasswordRotationTime()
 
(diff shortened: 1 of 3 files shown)


pkg/databaseclaim/databaseclaim.go
```diff
@@ -1216,27 +1220,32 @@ func (r *DatabaseClaimReconciler) manageUserAndExtensions(ctx context.Context, r
 		}
 	}
 
-	if status.UserUpdatedAt == nil || time.Since(status.UserUpdatedAt.Time) > rotationTime {
-		logger.V(1).Info("rotating_users", "rotationTime", rotationTime)
+	if statusNewDB.UserUpdatedAt == nil || time.Since(statusActiveDB.UserUpdatedAt.Time) >= rotationTime {
+		logger.V(1).Info("rotating user password",
```
drewwells, line 1224: querying logs without spaces is faster
```suggestion
		logger.V(1).Info("rotating_user_password",
```

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1162,10 +1162,14 @@ func (r *DatabaseClaimReconciler) createDatabaseAndExtensions(ctx context.Contex
...
+func (r *DatabaseClaimReconciler) manageUserAndExtensions(ctx context.Context, reqInfo *requestInfo, logger logr.Logger, dbClient dbclient.Clienter, statusNewDB *v1.Status, statusActiveDB *v1.Status, dbName string, baseUsername string, operationalMode ModeEnum) error {
 
-	if status == nil {
-		return fmt.Errorf("status is nil")
+	if statusNewDB == nil {
+		return fmt.Errorf("status newDB is nil")
+	}
+
+	if statusNewDB.ConnectionInfo == nil {
+		return fmt.Errorf("newDB connection info is nil")
```
drewwells, line 1172: StatusNewDB can be blank in many cases, this will prevent existing dbs from reconciling

pkg/databaseclaim/databaseclaim.go
```diff
@@ -1205,7 +1209,7 @@ func (r *DatabaseClaimReconciler) manageUserAndExtensions(ctx context.Context, r
 		if err := dbClient.UpdateUser(oldUsername+dbuser.SuffixA, dbu.GetUserA(), baseUsername, userPassword); err != nil {
 			return err
 		}
-		updateUserStatus(status, reqInfo, dbu.GetUserA(), userPassword)
+		updateUserStatus(statusNewDB, reqInfo, dbu.GetUserA(), userPassword)
```
drewwells, line 1212: This doesn't look right to me, we need to look at this more

drewwells (approved):

PR #423, opened 2025-06-16
PTEUDO-2835 - automation of grant role for new users as a part of KOP…

LenaRyazanov:
# Database Role Grant Script

A bash script for managing PostgreSQL role grants in Kubernetes environments using db-controller.

## Features

- Grant PostgreSQL roles through Kubernetes secrets
- Process multiple grants from a text file
- Check role membership before and after granting
- Detailed logging with timestamps
- Support for both single grants and batch processing


LenaRyazanov:
implemented under db-controller-scripts: [https://github.com/Infoblox-CTO/athena.devops/tree/main/scripts/db-controller-scripts](url)

